PR #64, opened 2017-06-23
Frontend

skept:
Relatively early working version of the frontend code with the following functionality:

- Analysis Builder interface with Overview and Predictions tabs
- Save and/or Generate the "draft" analysis
- Load previously saved analyses into Analysis Builder
- Analysis management interface to list, open, and clone saved analyses
- Client-based user sign-up and sign-in

See web/frontend/DEVNOTES.md for instructions on how to run the frontend development server


docker-compose.yml
```diff
@@ -8,9 +8,10 @@ services:
     links:
       - postgres:postgres
     volumes:
-      - /usr/src/app/static
+      - /usr/src/app/frontend/build
       - migrations:/migrations
       # ADD OTHERS VOLUMES HERE
+      - /Users/pouya/projects/neuro/shared_data:/shared_data
```
adelavega: This should probably not be commited. You can keep a local copy that doesn't get pushed with this info

adelavega:
Actually pouya I think you need to merge master first, because my latest changes will almost certainly break this (in particular, the status field, vs locked field)

skept:
@adelavega I took care of the items we discussed. Should be ready to merge.

adelavega:
Great! I'm deleting the frontend branch from now. If possible, lets isolate specific features in new branches and delete the branch after. Calling it "frontend" would be fine as long as we have a delineating between new features and merges. Additionally for features that require both frontend and backend changes we can work on the same branch.

Issue #65, opened 2017-06-23
Add lock route


adelavega:
Instead of monitoring status changes

Issue #63, opened 2017-06-23
Add additional fields to /api/user response


skept:
Currently the response type is 

```typescript
export interface ApiUser {
  email: string;
  name: string;
  analyses: { hash_id: string, name: string }[];
}
```

Would like it to be:

```typescript
export interface ApiUser {
  email: string;
  name: string;
  analyses: { hash_id: string, name: string, 
              status: string, description: string, modified_at: string }[];
}
```

We need this information for the main analysis management page (i.e. homepage). This will still be a very small payload but will save lots of additional roundtrips to the server and simplify the frontend code. See #62 about the new `status` field.

Issue #62, opened 2017-06-23
Add new status field to analysis model 


skept:
As we discussed in the specs doc, we'll need a new status field with possible values of DRAFT, PENDING and GENERATED. I think the existing boolean `locked` field will technically be redundant since it can be derived as `locked = status != 'DRAFT'`.

skept:
@adelavega I think it would be fine to keep the `locked` field (so we don't have to make too many changes to the existing code) and add a `status` field with the possible values of `PASSED`, `FAILED` and `PENDING`. Basically `status` would display the status of the latest submission. If the analysis fails, the locked flag should be cleared too so the client can make changes and resubmit.

Issue #59, opened 2017-06-22
Allow user to delete draft analyses


skept:
@tyarkoni @adelavega Do we want to support this?

tyarkoni:
You mean, any `Analysis` that hasn't been locked/compiled? Yeah, I think that makes sense.

skept:
Ok, I'll implement it in the frontend once the backend API is updated to support this.

PR #67, opened 2017-06-23
Backend adds

adelavega:
Closing several issues with this PR: #65, #63, #62, #59 & #53


web/resources/analysis.py
```diff
@@ -106,25 +107,46 @@ def get(self, analysis_id):
...
+		if analysis.status != 'DRAFT':
 			utils.abort(422, "Analysis is not editable. Try cloning it.")
-		elif 'locked' in kwargs:
-			if kwargs['locked'] > analysis.locked:
-				kwargs['locked_at'] = datetime.datetime.utcnow()
-				### Add other triggers here
 		return put_record(db.session, kwargs, analysis)
 
+	@doc(summary='Delete analysis.')
+	@utils.auth_required
```
tyarkoni: Many routes need to be restricted to ensure that users can only edit/delete their own analyses

skept:
@adelavega What are the possible values of the new `status` field? Do you still distinguish between `save` and `generate` by looking at a `locked` parameter passed to the API route or has that changed?

PR #73, opened 2017-06-24
Datalad & NeuroDebian support

adelavega:
- Add NeuroDebian to docker image
  - Needs work to be less repetitive, currently starting w/ python build. Might be better to start from NeuroDebian build, as that will be less individual package downloading, but need to properly test
- Add YAML configuration file. This describes datasets to be installed, and features to be extracted
- Support remote DataLad datasets. Uses AutomagicIO's patched open function to download resources as needed. Can be slow, try configuring w/ local copy of datasets as to reduce repetitive downloading.
    - Having a meta-dataset that can be maintained and stored remotely and locally would help with this as well, as well as keeping track of non-public (currently) stimuli.


adelavega:
This is not done (performance issues with neurodebian building) but let me know if the yml config seems ok to you @tyarkoni 
See: `web/confing/datasets.yml` for an example

A complimentary approach would be to define a datalad super dataset that includes all of the manually curated datasets we are interested in (including potential changes/forks required to make them compatible).

Issue #84, opened 2017-07-10
Add config json field to analysis


adelavega:
Unstructured array with misc fmri analysis config parameters

PR #85, opened 2017-07-10
Added config field, and nested transformation field

### LINKED ISSUE #83: Change analysis['transformations'] type to array
Currently the the API route to save the analysis expects an object for `transformations` and return a 422 response:

```javascript
{
  "message": {
    "transformations": [
      "Not a valid mapping type."
    ]
  }
}
```

adelavega:
To close #83 and #84 

#84 was straightforward, I added an unstructured JSON field for "config"

For #83, "transformations" field now takes an array of structure JSON  that conforms to the Transformation schema. I've also added a schema for "parameters". This might be overkill but should ensure appropriate inputs. 

My only question is if we should be referring to predictor names in inputs or predictor ids. The latter seems less error prone but is less human readable.

### DIFF
--- a/web/models/analysis.py
+++ b/web/models/analysis.py
@@ -22,7 +22,8 @@ class Analysis(db.Model):
     description = db.Column(db.Text)
     data = db.Column(JSONB, default={})
     filters = db.Column(JSONB) # List of filters used to select runs
-    transformations = db.Column(JSONB, default={})
+    transformations = db.Column(JSONB, default=[])
+    config = db.Column(JSONB, default={}) # fMRI analysis parameters
     created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow)
     modified_at = db.Column(db.DateTime, default=datetime.datetime.utcnow)
     saved_count = db.Column(db.Integer, default=0)
--- a/web/resources/analysis.py
+++ b/web/resources/analysis.py
@@ -82,6 +82,13 @@ class TransformationSchema(Schema):
 	inputs = fields.List(fields.Str(), required=True, description='Array of input predictors')
 	parameters = fields.Nested('ParameterSchema', many=True, description='Array of parameters.')
 
+	@validates('inputs')
+	def validate_preds(self, value):
+		try:
+			[Predictor.query.filter_by(id=int(r)).one() for r in value]
+		except:
+			raise ValidationError('Invalid predictor id in transformation input.')
+
 class ParameterSchema(Schema):
 	name = fields.Str(description='Parameter name.', required=True)
 	kind = fields.Str(description='Parameter type (boolean or predictor).',
--- a/web/resources/analysis.py
+++ b/web/resources/analysis.py
@@ -78,14 +78,15 @@ class AnalysisSchema(Schema):
 		strict = True
 
 class TransformationSchema(Schema):
-	name = fields.Str(description='Transformation name.')
-	inputs = fields.List(fields.Str(), description='Array of input predictors')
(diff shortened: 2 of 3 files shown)


adelavega:
image
This is what the analysis schema looks like in swagger right now.

adelavega:
image
This is what the example value looks like for analysis in swagger right now.

tyarkoni:
References should all be to numerical IDs. And I think `inputs` should be a list (also of numerical IDs) rather than a dictionary).

skept:
I'm also in favor of referencing predictor IDs instead of names since that's how I've already implemented it. The IDs are currently numerical but are passed between the client and server as strings, which I think is fine.

adelavega:
Actually for some reason that image was incorrect, and it does expect a list of strings, which are predictor ids as text, so its ready to merge

adelavega:
Actually, just to be sure I added validation of predictor inputs in transformations and updated the accompanying test.

adelavega:
@skept don't you think its a bit inconsistent to allow string predictor ids when they are integers elsewhere though? Ideally, it would just be an actual integer, and parameters with boolean value would be actual booleans not strings.

tyarkoni:
Yes, references between classes that are represented in the DB should always use integers (specifically, the ID columns in the DB tables).

adelavega:
Sounds good to me, but now for the field "value" in transformations, I need to define a custom Polymorphic field that can accept either booleans or integers, depending on the type. Looks doable though.

adelavega:
I added a custom polymorphic field that supports boolean and "predictor" types (aka integers that are validated to be existing predictors). Thus, transformations will actually look like:

    {"name" : "transformation_name",
     "value" : {"type" : "boolean", "value" : true}
    }

Look OK?

skept:
@adelavega I just double checked and the predictor IDs in the JSON payload (transmitted by either the client or the server) for all routes are indeed numbers (i.e. they're not wrapped in quotes). So ignore my previous comment. I just store them as strings in the UI state because it simplifies the UI code by eliminating the need for a bunch of `parseInt`s and `toString` calls.

Regarding the transformation schema, I'm a little confused. Is your example a transformation or a parameter within a transformation? I thought we settled on what I have below. I'm ok with changing 'kind' to 'type' if you prefer. I thought 'type' was a Typescript keyword so I went with 'kind' to but it looks like it does allow it as object keys even if it's not in quotes.

```javascript
[
  {
    name: 'standardize',
    inputs: [2, 6, 10, 26],
    parameters: [
      {
        name: 'demean',
        kind: 'boolean',
        value: false, 
      },
      {
        name: 'scale',
        kind: 'boolean',
        value: true, 
      },
  },
  {
    name: 'orthogonalize',
    inputs: [3, 4, 5],
    parameters: [
      {
        name: 'wrt',
        kind: 'predictors',
        value: [6, 7, 8], 
      }
    ]
  }
 ]
```

tyarkoni:
FWIW, I don't think the transformations schema needs to be encoded anywhere in the back-end at this point. The definitions of the transformations will probably be stored in a static json file that's kept in the client-side code. And what's returned from the client to the server after the user specifies transformations should--for the time being at least--be saved as a json object in the transformations field of `Analysis`. So I don't know that we need to enforce any structure on the transformations in the back-end just yet, @adelavega.

adelavega:
Yeah I think you're right @tyarkoni. It was just easy to do but now it seems to be causing problems. I'll just revert back to using a regular JSON field and @skept the example you posted looks correct to me.

Issue #47, opened 2017-05-19
Analysis bundle endpoint


adelavega:
Although I've left `predictor-events` as repetition json objects, for a final download from the `CLI`, a bundle which includes *all* dependencies makes sense.

adelavega:
UPDATE: Add an endpoint that downloads all files required for analysis (aside from original BIDS dataset).

tyarkoni:
What kind of dependencies? I think that, at least in the beginning, there shouldn't be anything to download on an analysis-by-analysis. I.e., all analyses created with the UI should have the same dependencies. So it would probably make sense to build all dependencies into the NeuroDebian and/or Docker images, so that all that's downloaded when a user executes an analysis on the CLI are the event files, model spec, etc.

adelavega:
I didn't mean library dependencies. I meant more like all of the configuration + event files + python script (maybe. to ensure same script that was used to compiled is used to run)

adelavega:
Closed with #113

Issue #46, opened 2017-05-19
Trigger creation of nipype workflows & analysis bundle


adelavega:
Upon locking of an analysis.
- Ensure these are async calls. 
-  Generate graphs  
- Add counts of workflow retrievals / downloads tables - timestamp every time someone retrieves a workflow.

adelavega:
Also create new event files (w/ only selected predictors).

adelavega:
Make sure file references are correct in db (i.e. to niftis in BIDS dataset, but also to newly created event files). This way grabbids is not necessary in the actual workflow as individual files are pointed to. If it turns out to be necessary (i.e. structural files?) that's fine.

qmac:
This issue was moved to PsychoinformaticsLab/neuroscout-cli#1

PR #113, opened 2017-07-31
Dockerize Celery / Trigger workflow generation

adelavega:
- Celery will be used for async jobs
- Celery and redis have their own docker images allowing for scaling (more redis images can be launced)
- When workflows are completed, graphs & workflows are available at new endpoints

To-do:
- Fix merge conflicts w/ master
- Trigger validation of workflows given frontend inputs (#46)
- Add analysis bundle endpoint (#47)


adelavega:
- Add tests

adelavega:
This PR also moved the main flask project to "neuroscout" as "web" was rather misleading. 
Also #126 will be done here

adelavega:
To-Do:
- Check/debug validation manually w/ downloaded .json
- Trigger actual validation in Celery (replace test code)
   - Should return an error code 
- Test triggering remotely

Ideally:
- Test running a simple workflow w/ bundle, and debug bundle

adelavega:
Celery worker is now triggered by the compile endpoint, but still need to debug the actual validation of the json, and send error codes back to the main app. 

Note: worker requires a lot of modules that are necessary for workflow. Might be good to set up a sub requirements file (or extras) for workflow, that worker also installs.

Tests may have to wait (or will fail on travis for a while) because they require the latest pybids, as well as nipype w/ BIDSTransformer module. Might have to pip install locally.

adelavega:
Ready to merge. To-do for another PR: 
- host celery tests on travis
- more complex tests of analyses. currently simple examples. probably should create an analysis generator function to make this easier (as currently requires lots of querying)
- add more comprehensive validation (e.g. applying transformation functions).

Issue #133, opened 2017-09-22
Ingest (scrape) from OpenNeuro


adelavega:
The goal when ingesting a dataset will be to have a full index of resources associated with a dataset, such that the CLI does not need do any indexing, only have a set of files required for an analysis (in JSON bundle) which are fed into DataLad for retrieval. 

Update population scripts so that datasets are added to index fmriprep files from openneuro. Need to scrape openneuro to find out the basepath of fmriprep executions (which are stored in config YAML files) and then index all of the subsequent files using Grabbit. Grabbit then generates an index of files and entities which are then ingested. 

Also: keep track of number of values in Run table. Should not need raw data after inital ingestion.

Also current problem: DataLad version of dataset generally don't have stimuli.

In the very future, it would be ideal to only keep track of entities (not paths), and DataLad should be able to index these.

adelavega:
TR also needs to be added to db, and bundle

adelavega:
For now ignoring grabbit index as that feature is not ready yet. Simply generating paths relatively to base path.

adelavega:
Closed with #136

Issue #132, opened 2017-09-22
Separate CLI/workflow into own repo


adelavega:
The CLI should be separated into its own github repo, made into a python package and docker image.

adelavega:
Would then also be good to move all issues with tag 'workflow' into that repo.

adelavega:
Actually just kidding this should be closed when this repo imports from other repo for any workflow related things.

adelavega:
Closed with #136

Issue #130, opened 2017-09-20
Apply/validate transformations server side, generate design matrix


adelavega:
When submitting an analysis for validation, transformations should be applied to event files to ensure they are valid.

adelavega:
All transformations should be applied in the validation/generation state and a full design matrix should be included in the JSON bundle.

adelavega:
Design matrix could be saved as file on disk or in server (JSON field)

adelavega:
Closed with #136

Issue #126, opened 2017-08-29
Ensure transformations in analysis (from front end) match BIDS model spec


adelavega:
"Input" instead of "Inputs", etc...

adelavega:
If possible, it may be necessary to have other references, but that would be ideal.

adelavega:
Related: Make name of predictor either string of int, or a more descritive name

adelavega:
Actually bundle will not have transformations, but perhaps the transformations from the front-end should respect BIDS model.

adelavega:
Duplicate

PR #136, opened 2017-09-28
Validate transformations & ingest from openneuro

adelavega:
WIP. This will address #126 #130 #132 and #133


adelavega:
#132 is done, workflows are being imported
#133 - Scraping from OpenNeuro is working, need to test against updated workflow. Simply using curated address in config file, and generating fmriprep relative paths. The goal will be to allow these paths to be using against remote or local files in CLI (for testing mainly). No need for DataLad right now really, simply fetch files in workflow to local folder, or point to local folder.

adelavega:
Merging without addressing #126. Will leave that to a front end PR.

PR #140, opened 2017-10-06
Add bundle fields / transformation tests

adelavega:
- Add preproc and dataset addresses


adelavega:
- Also adding transformations and contrasts to test analysis, and corresponding tests

Issue #142, opened 2017-10-06
Additional bundle endpoints


adelavega:
Currently, bundle is very verbose and user-unfriendly.

Add endpoints for json and tsv representations of the design matrix, and separate from main bundle endpoint.

Also, add filelist endpoint as json

adelavega:
Actually reorganize as follows:

- `analysis` endpoint returns full analysis spec, but with only ids for 
- `design_matrix` as both `/tsv` and `/json`. Store tsv on disk on compilation. Should not change after successful compilation. Returns `404` if not compiled.
- `images` returns list of preprocessed images required (relative path)
- `bundle` is tarball w/ all 3 (full analysis spec, design_matrix and image list)

Possible also include a `summary` human readable endpoint or a more comprehensive `analysis` end point with nested fields (e.g. `dataset.dataset_address`, `predictor.name`)

qmac:
Let's also keep https://docs.google.com/document/d/1bhFwSe4kqPsgqcjGp3XSM6Y1N2eAMi6xZrUp_c3mPGg/edit?ts=59ce704f in sync with this.

adelavega:
Closed

PR #144, opened 2017-10-06
Refactor analysis bundles, and add new analysis endpoints

adelavega:
WIP will address #142 

Uses Marshmallow schemas to create new "views" of the analysis. 

Still need to add file end points (such as tarball bundle and tsv events)


adelavega:
Need to update tests, also

qmac:
Can we do something more verbose like `full_analysis` instead of `full`?

adelavega:
Well it's nested under `analysis` so i think analysis is implied. Other more descriptive ideas?

qmac:
description? parameters? config? My main concern is that the file name once its retrieved is just `full` by default.

adelavega:
Hmm maybe `specification` but either way I don't think you should worry about it much because a) can't you save to a new file name e.g. `full_analysis` and b) soon the bundle will include the file so you don't need to fetch it. and I can call it `full_analysis` or `analysis_specification`

Issue #150, opened 2017-10-12
Last activity field should reflect last call from user


adelavega:
Everytime a protected resource is used w/ a valid JWT, record time in user record.

adelavega:
Closed w/ #151

Issue #79, opened 2017-07-06
Send verification/forgotten password emails to users


tyarkoni:
This may already be implemented, but in case it's not, we should make sure that users get emails to verify their accounts before they can start using them (and also allow password resetting). Rather than setting up a mail server, let's hook into a third-party service; we probably won't need to worry about paid plans for a long time (though we can pay when we need to). Mailjet or sendgrid should work well. Neurosynth hooks into Mailjet and that works fine (so you can look at how that's configured), but @skept uses sendgrid and says it works well too.

PR #151, opened 2017-10-12
Send email confirmation/verification/reset links to users

adelavega:
WIP. Resolves:
 - #79 - Sends emails to users 

To-do

- [x] - Update API endpoint for register to call auth register_user (to send email). 
- [x] - Work on confirmation template. Shows a status message (success, etc) and redirect link back to app. Ideally, this would be an api call with a redirect, but need to think about this more.
- [x] - Do not allow unconfirmed users to get token (update tests for this as well)
- [x] - Log when users last accessed protected route - #150 (e.g. `last_access` field)
- [x] - Implement password reset email request
- [x] - Refactor email to have reusable send mail function and templates (like flask-security)


adelavega:
Implemented the above. To get this production ready we still need to confirm the neuroscout domain, other we can only send emails to confirmed addresses using Mailgun sandbox. 

Also, for password reset, I implemented a token method, in which the token is given to the API as a POST. Presumably, in the future this would require the frontend to have a field to paste this token, or ensure the router is working correctly, and this can be made into a clickable link.

Issue #154, opened 2017-10-14
Send correct IP address / domain for email link


adelavega:
This should be a value set manually (e.g. domain) that it uses to generate the link.

PR #155, opened 2017-10-16
ENH: Set remote flask host & login frontend tweaks

adelavega:
- Among other things, setting remote flask host allows the confirmation emails to go to the correct email address.
- Also added a resend confirmation link endpoint.


adelavega:
This PR resolves:
#161 
#160 
#154 
#156

Issue #168, opened 2017-10-28
Refactor predictor to mirror extractor attributes


adelavega:
Many predictor attributes are direct mirrors of extracted feature attributes, which are copied over.

It might be beneficial to instead have dynamic properties linked directly, so no manual updating in necessary, and to ensure data does not diverge.

adelavega:
Actually, this can be done on the level of the output Schema, as extractor properties can be included as a NestedSchema, if they exist.

adelavega:
Tal thinks that this should not be a problem since this is what db's are design to do (hold large amounts of data). So I'm going to close this for now, unless it becomes an issue again.

Issue #162, opened 2017-10-23
Add feature versioning & periodic re-extraction


adelavega:
Currently, only a single version of each feature is stored. We need to handle incremental updating of pliers features.

First, pliers will soon support extractor version numbers, and extracts should include that information.

Second, we need a version for the features based on the output. That is, if extractors are outputting new values, we need to re-extract and store new version. 
In order to detect this, create a utility with a battery of ~100 stimuli against which APIs are tested. This utility will be run periodically as a cronjob, and results will be stored for features + stimuli. If results change (perhaps by a diff threshold), then neuroscout features will be re-extracted for the features that changed, and a new version will be assigned.

API: Only the latest version of the feature is output by the default, but there should be a way to access older version of the features (e.g. `/feature/name/[version]`?).

PR #169, opened 2017-10-30
Versioning

adelavega:
Addresses #168 and #162 
To-do
- [x] Ingest pliers versions
- [x] Use nested schemas to display extractor properties in predictors
- [x] Increment versions of extracted features when re-extracting, instead of replacing or ignoring
- [x] Show latest version with API, but have end points for previous versions
- [x] Create tools to detect changes in remote extraction APIs (w/ battery of ~100 stims)
- [x] Add cronjob to detect changes, and re-extract (subset) as needed.


adelavega:
Need to decide how sensitive we want to updater to be. Perhaps for numbers it should be a rounding value set as a parameter, and for labels it should be exact.

PR #172, opened 2017-11-11
Extract using Google APIs & generate analyses

adelavega:
Goal is to extract using Google APIs, and appropriate ingest into db (including nas), and to specify analysis using these predictors.

- [x] Dynamically set up keys & set up on server
- [ ]  Update README to reflect above
- [x] Deal w/ NAs
- [x] Compile a valid looking analysis


adelavega:
Current problem: All predictors from Google get loaded, but when loading them into an analysis and compiling the following error occurs:
`KeyError('Level predictor_id must be same as name (None)'`

Also ingest process never seemed to exit, so ensure that this is going OK and not crashing. It's definitely very slow so batch processing may be appropriate.

Issue #177, opened 2017-11-15
Save converted stimuli to db


adelavega:
Goal: To cache intermediate extractor results that result in intermediate stimuli being generated (e.g. re transcribing speech when only last part of node needs to change)

Rather than providing a full Graph to extract, explicitly provide converters (stim -> stim) and extractors (stim -> value) for each dataset. extractors are applied to all applicable stimuli in dataset, except those that are "inactive". 

For example, for german would omit transcription steps, and no implicit conversion would happen.

adelavega:
First, work on adding stimuli conversion step, and applying all extractors to all elligible stimuli for dataset/task. 

Maybe take out stimuli extraction from ingest into another function, which takes an optional schema of modiality to inactive inactive.

Next, add a way to set active/inactive for certain stimuli (so they would not get processed in next step).

PR #178, opened 2017-11-22
Explicit converters

adelavega:
Addresses #177 

- [x] Extract directly from stimuli in database + tests
- [x] Add converter extraction
- [x] Add tests
- [x] Renable feature updating, checking both extractors and converters


adelavega:
This may require a refactor, so that convertion and extraction is applied across *all* datasets. It's probably okay right now, but shared Stimuli is causing problems, because they could be shared across datasets. It's becoming more likely because dervied stims (e.g. text) may be shared.

Another potential solution: add some uuids to the stimulus hash, so that they are unique to a task. This probably works in the short term, but it would require essentially reextracting the same features twice in some cases.

adelavega:
Current problem is when I have onset as primary_key of RunStimulus, I get a ObjectDeletedError error late on. No idea why, but it's fixed by rexecting the query. But that is still too strange.

adelavega:
Trying to find a way to not use 'onset' (float) as primary key. Somehow that triggers the odd error.

adelavega:
Talked to Tal. Decision:

- [ ] Make stimuli specific to a dataset (so no worries on re-extracting, or sharing resouces)
- [ ] Make hash field not unique (because in theory it could overlap w/ the same stimulus in another dataset)
- [ ] Use is as primary key in RS table
- [ ] Add unique constraint on properties (onset, duration, foreign keys)

adelavega:
Only things missing are:

- [ ] Tests for updating functions (that only the correct subset changes when updated)
- [ ] Way to disable original stimuli from further processing on ingestion.

adelavega:
I think it's better instead to disable certain transformers for specific datasets. Or rather, in the future to define a mass set of transformers (plus parameters) and apply to all, except for the disabled ones. I'll leave that for another PR.

Issue #180, opened 2017-12-05
Create master pliers transformer list


adelavega:
Since transformers often have to parameterized, a master list would be useful. Within datasets, specific transformers can be disabled or the set can be entirely overwritten.

adelavega:
Maybe instead extract this from feature schema? And can remove add on individual level?

Issue #176, opened 2017-11-15
Add feature post processing


adelavega:
When features are ingest, we may want to process the data before and after insertion to the database. We can do this via a series of callbacks that map onto features, and return a transformed value. 

`extractor_preprocessing` - functions to before putting in the data into db, such as transforming from string to integer, etc..

`extractor_postprocessing` - functions to apply afterwards to create new feature. This could include derived features that take multiple features as inputs (e.g. number of faces). 

Input field should be regex to allow multiple fields to be input.

Issue #175, opened 2017-11-14
Feature schemas: match names using regex


adelavega:
Some APIs may return undetermined output, so use regex to match & rename feature names

Issue #174, opened 2017-11-14
Feature schemas for APIs


adelavega:
In #157 we addressed feature schema implementation. Now, a master schema that spans all APIs we use is needed.

We might want to look into adding post processing, for predictors such as Google Face probability api which returns strings, not numbers.

PR #182, opened 2017-12-09
Master schemas

adelavega:
Resolves ingestion issues:

- [x] #174 create master feature schemas
- [x] #175 support regex
- [x] new regex tests
- [x] #180 create master transformer list
- [ ] #176 - add feature pre and post processing


adelavega:
For feature preprocessing, include in feature schema.
For post-processing (e.g. functions that create new predictors from old ones), could either apply at the whole database level (presuming the same type of predictors will always want to be transformed), or could include references so functions on a dataset by dataset level in the main dataset.json file.

Most likely approach, define list of operations for whole database, but can opt out at the dataset level with dataset.json, sort of a combination of both approaches, and similar how you can opt out of default converters, etc...

adelavega:
Leaving #176 for later

PR #183, opened 2018-01-04
Feature pre/postprocessing transformations

adelavega:
Addresses #176 

Preprocessing is applied at the feature schema level (prior to inserting in db). Potentially unnecessary if we could just keep features hidden, but already implemented in `master.`

Postprocessing can take 1 or more features and generates a new feature.


adelavega:
@tyarkoni if you get some time, let me know if you have any comments. Main thing is I realized executing more complex transformations is a bit tricky in SQL. (e.g. how many colors (from GooglePropertyExtractor), does a stim have over a certain threshold). The queries were just getting slow, complicated and error prone. 

Do you think these sort of "transformations" are worth doing up front, or should we stick to simple ones (e.g. number of faces). Any other you can think of for now?

adelavega:
(The threshold one that I have in there right now is not working, because the values are represented as stings, and casting fails sometimes because values are sometimes not castable (e.g "VERY_LIKELY").)

I suppose that would be another useful transformation, to make all qualitative values such as those into numbers. To minimize the transformations users have to apply on their end.

adelavega:
Hmm... Yeah I assumed most transformations would be simple enough to do in SQL, and it is faster. 

But I suppose the framework I have doesn't require them to be done in SQL, as the transformation functions just return a list of dicts to populate new `ExtractedEvents` (and the `apply_transformations` function create those plus a new `ExtractedFeature`). That might work nicely, because for simple things such as counting, I can easily do that in SQL, but I could always dump to pandas for a more complex function.

Regarding doing most at insertion, I originally was doing that, but then I realized there will be many times we will change our mind or come up with new transformations, so I prefer to do it from  "deactivated" features for the most part. The current set up finds the most recent version of each feature queried (for that Dataset/Task), and passes the subquery to the transformation function.

adelavega:
Also in this PR:

- [ ] Fix auto-model run_num bug
- [ ] Fix text ingestion hashing bug, and make ingestion more efficient

adelavega:
@tyarkoni I ended up re-factoring such that db contents (`ExtractedEvents` for the given runs) get put into a pandas df, and transformations are applied there. Have tested it, and the final predictors coming out of it (`num_faces`, `num_colors`, etc...) and it looks fine. Also fixed a bunch of random bugs.

I'm going to merge it unless you have any other thoughts. Tests are just waiting on the pliers PR I submitted.

Issue #185, opened 2018-01-09
Switch celery compile to use pybids analysis instead of transformer module


adelavega:
Use the Analysis module instead of deprecated Transformation module

PR #186, opened 2018-01-09
Fix preprocessing, update pybids, pliers

adelavega:
This PR fixes some minor bugs with preprocessing previously implemented, including updating pliers. 

Also, updates to use pybids analysis module when compiling analyses (#185 )


adelavega:
Left updating pybids to another PR

Issue #207, opened 2018-04-08
Feature schema cannot have more than one key name


adelavega:
Feature schema is invalid JSON, because multiple extractors have same key name. Instead, need to use an array, with objects, with a "name" attribute so that they dont clash....

For example, ClarifaiAPI

adelavega:
Actually, just using the object's as an array of objects might work...

Issue #205, opened 2018-03-26
Feature schemas - object id


adelavega:
Pliers now returns an object id, instead of prepending feature names with face id. Update feature schema accordingly.

adelavega:
Currently, just adding new instances into db without distinguishing them.

adelavega:
Add object ID column to query

Issue #111, opened 2017-07-31
Ingesting datasets is slow


adelavega:
Looping over event files and extracted events (especially for movies) is *very* slow. Profile performance (might be SQL or just looping problem) and optimize ingestion functions.

PR #209, opened 2018-04-09
Fix/feature schema

adelavega:
#207, #205, #210 & #111

### DIFF
--- a/celery_worker/tasks.py
+++ b/celery_worker/tasks.py
@@ -5,7 +5,6 @@ import tarfile
 import json
 from pathlib import Path
 from tempfile import mkdtemp
-from bids.variables import load_variables
 from bids.analysis import Analysis
 from bids.grabbids import BIDSLayout
 from celery.contrib import rdb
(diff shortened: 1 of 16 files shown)


adelavega:
This will also try to address #111 to some extent by adding a few bulk inserts

adelavega:
Things are looking good, ingestion is taking a lot less time, and it looks to be outputting reasonable event files that fitlins can fit (currently running: model w/ 10 Clarifai + 2 low level features on Generic Object Decoding, on around 25 runs on 5 subjects). 

@tyarkoni I'm merging this but FYI the main issue I was having is creating `Predictor` tables from `ExtractedFeatures` was going horribly slow (e.g. 8 hrs). I re-wrote that part of the code, using a lot more bulk inserts and avoiding as many expensive queries as possible in loops, and I think it's down quite a bit. In the long run, it would make sense to "mirror" EFs rather than copy them over, as that can be quite wasteful.

adelavega:
One remaining question: Why some features were marked as "inactive" at first... Test with small amt of features in subset.

Issue #164, opened 2017-10-23
Environment variable based hostname


adelavega:
Currently, the hostname is hardcoded to the frontend. This value should come from an environment variable at build time (yarn build), or in some other way that allows it to be dynamic for test (e.g. localhost) and deployment scenarios.

PR #229, opened 2018-05-08
224 fetch run predictors and more.

### LINKED ISSUE #224: Fetch only predictors associated with runs
It seems as though the Predictor tab is listing all predictors, not just the ones for that run. The API seems fine, so this must be a frontend bug. This makes it hard to select the correct predictors as there are so many.

rwblair:
Fixes #224 and potentially #164 

The server url used in the front end was not put into an environment variable but is in a config file. 

Also updates the nginx image being used by docker since the current one in master is no longer supported.


adelavega:
At first glance, changes LGTM. I need to clean up a few things on the testing/production server, and then im going to try running this branch there to give it a whirl.

PR #232, opened 2018-05-10
Update front end packages

rwblair:
Some of the packages were over a year out of date. Incidentally the update from antd 2.x to 3.x makes the default language English, so now tables with nothing in them say 'no data' instead of 暂无数据.


adelavega:
Awesome. Looks like the frontend build is no longer compiling, though. Probably would be good to thoroughly test this, as I expect some downstream bugs from so much upgrading. I'll miss the Chinese... It was "quirky".

rwblair:
Yeah once I removed my local node_modules because Travis failed my local deploy stopped installing. So strange, had the tests passing and the site working but once I rm'd node_modules and tried to reinstall everything broke bad.

adelavega:
Yeah, not surprised, but good thing to do anyways!

adelavega:
hey, hows this coming along? looks like tests are passing now

adelavega:
Allright, looks good merging!

PR #242, opened 2018-05-17
Background SQL queries

### LINKED ISSUE #239: Compile route timeout
It seems that if a response takes too long, flask will kill the worker. In this case, if there are too many predictors, the `/compile` route will take too long, and timeout. It's also annoying how long this takes, so we need to work on a way to speed up the queries used in `compile` or to push them to a background job, and return `200` immediately when a compile request has been made.

adelavega:
Ideally, SQL queries can be run in the background. In this PR, I try to push the flask context to celery, so that it can access the DB. Fixes #239


adelavega:
Allright this re-organizaton is turning out to be a mess because of circular imports. I think the easier solution will be to *not* import app from the `neuroscout` (in celery), and instead launch a simple Flask and db instance that get configured over there, using the appropriate config (only aspect that is imported, could point at a file).

adelavega:
Will start over with a lightweight Flask instance in Celery.

tyarkoni:
Oh, sorry, I think I misunderstood: I'm not currently importing the app from flask; separating the celery tasks from the main app does seem like the way to go. I thought you were talking about how to get around having to install all of the same dependencies--for that I think it's probably easier to re-use the flask container than to create a new one just for the workers.

adelavega:
Yeah the issue is importing the app from flask-- circular dependency hell.

PR #244, opened 2018-05-22
Add shot change feature, life transcript

adelavega:
Adds: Google Video shot change feature support (transformation required), and fixes bug relating to using a transcript for multiple video files (Life dataset).


adelavega:
Remaining issues: shot change transformation not working as intended.

Issue #225, opened 2018-04-27
Specify a full BIDS model. Move transformations and contrasts to model. 


adelavega:
This will actually be a fairly major one, since previously the assumption was that the backend would construct the model given the right fields. But now, we will construct the whole model on the frontend.

This is a good idea for various reasons, although its more work on the frontend. One of them is that it decouples the frontend from Neuroscout. Potentially the logic user here, if sufficiently modular, could be used in a similar app that helps users make models of local datasets.

There are a few things that need to get done:

- [ ] Move transformations and contrasts into the model
- [ ] List the names of the predictors selected in "HRF_variables" and "variables" (these are the names of the predictors selected by the user *prior* to any transformations.  These get lised at "level" = run and dataset.
- [ ] Fill in "inputs". The "inputs" field is the set of unique "entities" that identify the runs. I would come up with the list by inspecting the fields "number" "session" "subject" in the selected runs, and idetifying the unique set of these items. In addition "task" should be included in inputs as well.

adelavega:
Here's an example of what a valid POST/PUT request without those fields (and the information in "model") looks like:

```
{
  "TR": 1.5,
  "compiled_at": null,
  "created_at": "2018-04-20T22:0",
  "data": {},
  "dataset_id": 5,
  "description": "",
  "hash_id": "dz",
  "model": {
    "blocks": [
      {
        "level": "run",
        "model": {
          "HRF_variables": [
            "brightness",
            "vibrance",
            "sharpness"
          ],
          "variables": [
            "brightness",
            "vibrance",
            "sharpness"
          ]
        },
        "transformations": [
          {
            "input": [
              "brightness",
              "vibrance",
              "sharpness"
            ],
            "name": "scale"
          }
        ]
      },
      {
        "level": "dataset",
        "model": {
          "variables": [
            "brightness",
            "vibrance",
            "sharpness"
          ]
        }
      }
    ],
    "description": "",
    "input": {
      "subject": [
        "24",
        "34"
      ],
      "task": "MerlinMovie"
    },
    "name": "MerlinCompareGOD"
  },
  "modified_at": "2018-04-20T22:0",
  "name": "MerlinCompareGOD",
  "parent_id": "V1",
  "predictions": "",
  "predictors": [
    {
      "id": 351
    },
    {
      "id": 352
    },
    {
      "id": 353
    }
  ],
  "private": true,
  "runs": [
    {
      "id": 134
    },
    {
      "id": 117
    }
  ],
[...]

PR #245, opened 2018-05-23
[WIP] frontend transformations

rwblair:
This addresses the three points in #225.

One outstanding issue is that the predictors are being stored in the model  using their unique IDs from the database and not their full names as shown in the spec. The work I did on contrasts and hrf_variables in this PR save them using their full name instead of ID. Predictors selected in transformations are still using their database IDs instead of name, will fix that next. Here is a sample of what the current model json being saved in the database looks like:

```
{
  "name": "test name",
  "input": {
    "exclude": {
      "subject": [
        "01"
      ]
    },
    "include": {
      "run": [
        "2"
      ],
      "task": "emotionalregulation"
    }
  },
  "blocks": [
    {
      "level": "run",
      "model": {
        "variables": [
          1,
          2
        ],
        "hrf_variables": [
          "trial_num"
        ]
      },
      "contrasts": [
        {
          "name": "test cont",
          "weights": [
            1
          ],
          "predictors": [
            "trial_num"
          ],
          "contrastType": "T"
        }
      ],
      "transformations": [
        {
          "wrt": [
            2
          ],
          "name": "orthogonalize",
          "input": [
            1
          ]
        }
      ]
    }
  ],
  "description": "test desc"
}
```


adelavega:
Cool, looks good to me so far. By the way, I probably can't test this until next week. On that topic, it would probably be a good idea to have a staging server so we can test there before going "live" as I currently do. Not that big of a deal right now, but it will be in the future.

rwblair:
I can setup a testing server on a spare server we have in the lab (when @chrisfilo is done with afterhours) or on AWS whenever we ready for it.

With the most recent pushes the model in the database should use predictor names instead of Ids everywhere so Im going to pull the wip tag off :

```
{
  "name": "test name",
  "input": {
    "exclude": {
      "subject": [
        "01"
      ]
    },
    "include": {
      "run": [
        "2"
      ],
      "task": "emotionalregulation"
    }
  },
  "blocks": [
    {
      "level": "run",
      "model": {
        "variables": [
          "trial_num",
          "reaction_time"
        ],
        "hrf_variables": [
          "trial_num"
        ]
      },
      "contrasts": [
        {
          "name": "test cont",
          "weights": [
            1
          ],
          "predictors": [
            "trial_num"
          ],
          "contrastType": "T"
        }
      ],
      "transformations": [
        {
          "wrt": [
            "reaction_time"
          ],
          "name": "orthogonalize",
          "input": [
            "trial_num"
          ]
        },
        {
          "name": "scale",
          "input": [
            "trial_num"
          ],
          "demean": false,
          "rescale": true
        },
        {
          "name": "scale",
          "input": [
            "trial_num"
          ],
          "demean": false,
          "rescale": true
        }
      ]
    }
  ],
  "description": "test desc"
}

**Comment by chrisfilo** [comment]
In the interest of time I would recommend doing this on AWS.

**Comment by adelavega** [comment]
OK cool. Well for now its fine as we're still in alpha and we have no users. I should be able to review/merge tomorrow or early next week. But it looks good and I don't foresee too many problems.

**Comment by adelavega** [comment]
Although I think we may need to discuss the layout in more detail later (I think it would be more intuitive to separate confounds from non-confounds in the list, and assume the latter won't be HRF variables), for now let's go with this design.

There were a few issues though:

1) Although the spec says that `input` should have an `include` and `exclude` section, in practice we have been using it this way in pybids. That's probably a bug with pybids now that I think about it, but for now, `input` should just take what would go in `include`. 
2) Input should take the "set" of runs, subjects, sessions, rather than a concatenated list of runs. E.g. if 10 runs are included, and 5 are run 1 and 5 are run 2, it should only say "run": [1, 2].
3) Also include the set of subjects and sessions. Currently, only run numbers are listed. 
4). If there is no run id (which is sometimes true, when there is only a single run), exclude that. See:

E.g.: 
screenshot from 2018-06-04 15-57-22

This also applies to sessions (most of the time there will not be a session)

5) If I add a transformation, and click "save and generate", but don't click "save" before, it wont be added to the model. We can either make "save and generate" call PUT, then compile, or rename the button to only ever say "Generate" and be greyed out if there are unsaved changes.

**Comment by rwblair** [comment]
Actually, could you test the 'save & compile' feature to confirm it goes off like we want?

Other than that should be good to go.

**Comment by adelavega** [comment]
I did test it and that looks good. *but* actually one more thing is required for these models to run. We need a "dataset" level block. 

See, for an email of a valid model w/ a dataset level block:

```
{
  "name": "RaidersVisualAuditory",
  "input": {
    "run": [
      "4",
      "1",
      "5",
      "8",
      "2",
      "3",
      "6",
      "7"
    ],
    "task": "raiders",
    "subject": [
      "rid000005",
      "rid000011",
      "rid000014",
      "rid000015",
      "rid000020",
      "rid000028",
      "rid000029",
      "rid000033",
      "rid000038",
      "rid000042",
      "rid000043"
    ]
  },
  "blocks": [
    {
      "level": "run",
      "model": {
        "variables": [
          "brightness",
          "vibrance",
          "sharpness",
          "rmse",
          "face_detectionConfidence",
          "architecture",
          "daylight",
          "indoors",
          "nature",
          "pattern",
          "room",
          "street",
          "technology",
          "vehicle",
          "writing",
          "speech",
          "shot_change",
          "num_colors_0_07",
          "aCompCor00",
          "aCompCor01",
          "aCompCor02",
          "aCompCor03",
          "aCompCor04",
          "aCompCor05",
          "X",
          "Y",
          "Z",
          "RotX",
          "RotY",
          "RotZ"
        ],
        "HRF_variables": [
          "brightness",
          "vibrance",
          "sharpness",
          "rmse",
          "face_detectionConfidence",
          "architecture",
          "daylight",
          "indoors",
          "nature",
          "pattern",
          "room",
          "street",
          "technology",
          "vehicle",
          "writing",
          "speech",
          "shot_change",
          "num_colors_0_07"
        ]
      },
      "contrasts": [],
      "transformations": []
    },
    {
      "level": "dataset",
      "model": {
        "variables": [
          "vibrance",
          "sharpness",
          "architecture",
          "daylight",
          "indoors",
          "nature",
          "pattern",
          "room",
          "street",
          "technology",
          "vehicle",
          "writing",
          "face_detectionConfidence",
          "brightness",
          "num_colors_0_07",
          "rmse",
          "shot_change",
          "speech"
        ]
      }
    }
  ],
  "description": ""
}
```

**Comment by adelavega** [comment]
Hmm, actually, this is a bit tricky because the list of variabes should actually match to the names of the first level contrasts. Currently, if there are no contrasts, all the "HRF_variables" get their own contrast made automatically. But actually, this should only be true if "auto_contrasts": True. 

Maybe we should leave this to another PR.

**Comment by tyarkoni** [comment]
Technically I think the spec requires that all levels before the last be present even if they're empty except for `'auto_contrasts = true'` (to keep propagating contrasts defined at previous levels). Also, we probably don't want to put anything in the `model` section past the first level; for probably all supported models (at least initially), everything we need will go in `contrasts`.

**Comment by adelavega** [comment]
Also: correction, "Save & Generate" does not seem to be working...

**Comment by adelavega** [comment]
Is auto_contrasts also allowe on the `dataset` level? If so would the following be all we need:

```
 {
      "level": "dataset",
      "auto_contrasts": true
}
```

**Comment by tyarkoni** [comment]
Yep, I believe that should be sufficient. Not sure if fitlins can handle that yet, and pybids also needs to be updated to handle auto_contrasts, but I think that's what's in the spec

**Comment by adelavega** [comment]
I can confirm that it can't. Let's try to meet the spec in Neuroscout, then.

Well, in this case then, @tyarkoni can you confirm that `input` should have `include` and `exclude` sections? Because currently in pybids it's not implemented like that (oops, maybe it was me). 

If so, @rwblair maybe we should revert that change then. And we can fix it in pybids soon.

**Comment by rwblair** [comment]
I didn't take away from the spec that all levels up to dataset are mandatory so that's good to know. I'll generate the subject and dataset levels with auto_contrasts.

Using `include` and `exclude` is specified in the spec as something that _can_ be done but is not mandatory. I can add the include and exclude keys back as optional fields to the input type in coretypes.ts.

**Comment by rwblair** [comment]
Right now if we have the following 4 files:
'''
sub-01_run-01
sub-01_run-02
sub-02_run-01
sub-02_run-02
'''

The user could select just `sub-01_run-01` and `sub-02_run-02` as inputs. I'm not seeing how just these 2 inputs could be represented with `include` and `exclude`.

Right now we are inclusive and say:
```
include {
  run: [1, 2],
  sub: [1, 2]
}
```

Is the current way these inputs are set in the model problematic?

adelavega:
I think you're right about all that. I think those are optional, and I also think they underspecify that situation... As far as we're concerned, it doesn't matter *too* much since we have a record the specific runs used in each analysis (and onlyt hose event files are generated in the bundle), but it does seem like a limitation of the spec.

tyarkoni:
Good point. I'll have to think about this some more. But it's possible we'll end up leaving things as is in the spec, because allowing include and exclude directives to be arbitrary regexes on the full paths seems like a recipe for trouble. Possibly the way we should handle this is to allow the user to specify a full list of input images in addition to (it instead of) the BIDS entities. That would work for our purposes, and offload any regexing onto the person creating the model spec rather than the receiving software.

tyarkoni:
@rwblair, re: it not being clear all levels are mandatory, thanks, that's good to know--I'll try to make that clearer. Personally I'd actually prefer to have automatic propagation across levels, but I've been convinced by other folks like JB Poline that in this case, explicit is better here. Also, if we propagate automatically, that means we'd be making `auto_contrasts = True` the default for all levels, which doesn't seem desirable. In any case, we definitely need more worked examples in the spec that illustrate this principle.

tyarkoni:
@adelavega, yeah, pybids is currently missing some stuff when reading in model specs, including the filtering directives. This will be a good hackathon task for me.

adelavega:
Hey one more thing @rwblair, the type of runs is always int. Currently they string in quotes, but they need to be casted in the model json field.

Thanks

adelavega:
Cool, so what's missing @rwblair? Just updating the model so it conforms to the standard exactly?

adelavega:
@rwblair any chance that we can add the changes that are missing (e.g. making a full model including the 'dataset' level) by monday? If so we might be able to launch alpha before my talk.

Issue #270, opened 2018-07-09
Make analysis table sortable on all headers


rwblair:
Default to sort by date

rwblair:
fixed in #271

PR #271, opened 2018-07-10
270 analysis sort

rwblair:
Forked from #245 should wait for that to be merged. Only commit(s) of interest are from
3ccab09 on.


rwblair:
fixes #270

PR #278, opened 2018-07-11
Fixes bugs in prediction selection.

rwblair:
Fixes bug when filtering and selecting a new predictor all the unseen but selected predictors would be dropped.

Fixes issues with removing predictors from side bar not being properly removed from table.

Fixes removing predictor from main list not also being removed from selected hrf predictors.


adelavega:
Ah, looks ike putting the frontend config file in .gitignore is causing tests to fail.

rwblair:
Might check out styling changes from that last commit. Most other tables in the project are using pagination so it brings the predictors in line with that.

The header size changes from page to page. If we don't like that I can try and find good fixed values for column width.

PR #279, opened 2018-07-12
Add "submitting" status, to workaround slow compile queries

adelavega:
This PR attempts to work around #239, and the generally slowness of submitting an analysis for compilation.

The issue is that the query is slow to get the PredictorEvents required for writing out bundle / compiling. We could run this query in Celery, but setting up Celery to be aware of SQLAlchemy, Flask + models is a pain, and requires a lot of restructuring.

An easy solution, is to simply have a status "Submitting" which is true after a user has clicked compile, but before the query has finished and the task has been submitted to Celery.

This way, the frontend can just stop caring about the response to `/compile` and instead just poll `/status` to see how the analysis is getting on. This may cause problems with scalability, but we can deal with that problem later, if we are that successful.

### DIFF
--- a/neuroscout/models/analysis.py
+++ b/neuroscout/models/analysis.py
@@ -37,7 +37,7 @@ class Analysis(db.Model):
 
     status = db.Column(db.Text, default='DRAFT')
     __table_args__ = (
-    	db.CheckConstraint(status.in_(['PASSED', 'FAILED', 'PENDING', 'DRAFT'])), )
+    	db.CheckConstraint(status.in_(['PASSED', 'FAILED', 'SUBMITTING', 'PENDING', 'DRAFT'])), )
 
     compile_traceback = db.Column(db.Text, default='')
     celery_id = db.Column(db.Text) # Celery task id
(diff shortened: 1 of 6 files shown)


tyarkoni:
This might be a case where it's best to execute an optimized SQL directly rather than dealing with the ORM overhead. I've run into cases in the past where the naive SQLAlchemy approach is basically unworkable, but splitting things up into a couple of queries or using raw SQL does the trick without any noticeable slowdown. Can you point me to the query that's causing the problems?

adelavega:
That's a possible solution. Actually in that scenario, even if the query is a bit slow, we could execute it in Celery as it would be much easier to implement SQL without the ORM.

adelavega:
its this function: https://github.com/neuroscout/neuroscout/blob/e7350c21abf390655526abba7834db7f67ff79a3/neuroscout/resources/analysis/endpoints.py#L79

adelavega:
(and I don't think its the iteration I'm doing in Python prior to the SQL call btw)

adelavega:
I added indexes (including a composite index) to PredictorEvent. This significantly sped up queries. Now the bottle neck seems to be the ORM overhead. I think queries are at a reasonable clip though (~22s for an analysis with ~62k PredictorEvent rows, ~12 predictors & 88 runs). Which ever way we choose to serialize those PredictorEvents will take time.

These indexes also sped up querying `/api/predictors`  when filtering by runs. However, the join with `PredictorEvent` (to be able to filter by run_id) is still the bottle neck (~0.5-1s). If this feels too slow, I added the ability to query `/api/predictors` by dataset_id. This will not guarantee the correct set of Predictors, but I believe that for most datasets with only one task, this will work.

PR #282, opened 2018-07-13
Fix/confounds

### LINKED ISSUE #273: Ingested confounds are written out incorrect in bundle
The values are wrapped in a list, and thus are not read in correct by pybids

adelavega:
Fixes #273 
I also used psql to fix incorrect values in the database

### DIFF
--- a/neuroscout/populate/ingest.py
+++ b/neuroscout/populate/ingest.py
@@ -46,16 +46,10 @@ def add_predictor(predictor_name, dataset_id, run_id, onsets, durations, values,
 
     values = pd.Series(values, dtype='object')
     # Insert each row of Predictor as PredictorEvent
-    pes = []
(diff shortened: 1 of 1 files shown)


tyarkoni:
That `.values.values` thing is kind of confusing... I should probably rename it to something better.

adelavega:
Yeah. Well blame pandas. But actually the issue for me was that  Dense variables return a dataframe (with a column named the variable name) and Sparse ones return a Series (I believe). So asking for `values.tolist()` of a dataframa will wrap values in a list (because its a list of column values)

tyarkoni:
Oh, right. We could potentially change that to be consistent, but I think there was a reason for it I'm forgetting. Anyway, as long as it works now, it's probably fine.

Issue #115, opened 2017-08-02
Poll API for updated analysis status


adelavega:
Doesn't look like the analysis status is being updated correctly. Only when I click "edit" from the homepage will the analysis update I believe.

It should probably check every x seconds, or everytime a tab changed if the status is "PENDING"

skept:
Yeah, it currently doesn't poll the server to see if the analysis is still "Pending". It's a good idea though. Should we be doing this on both the Analysis Builder page and homepage (which would normally have a list of analyses)? How long does it typically take for the analysis run to complete?

adelavega:
I'm still working on compiling... and that may vary. For now it's simple so maybe a minute but in the future we may do heavier validation. Everytime you do an API call to analysis, I update that field, FYI, so need to check a special "status" route.

skept:
Ok, related to this, have you also figured out what the server will return as the analysis results? Perhaps for now you can just return a JSON field that I'll just blindly display in the review or status tab? Once I know this I can implement polling and updating the status/results.

adelavega:
No all we need is to tell the user it is compiled. It means they can now go run the actual analysis on their computer, so we will add those instructions later. at /api/analysis/{id}/bundle the full analysis can be fetched by the command line interface, so compiling makes that end point available, but I don't think you need to worry about it aside from alerting the user to the status for now.

skept:
Ok, that makes more sense now. I thought the analysis would automatically run on the server. Do you already have a route for just querying the status? If not, I think at least for now it should be fine to just poll `/api/analyses/{analysis_id}`. The payload is pretty small and it happens in the background so a slightly faster route is unlikely to increase the perceived speed of the UI.

adelavega:
Yeah that works for now. If we hit performance issues we can change approach.

rwblair:
fixed in #283

PR #283, opened 2018-07-13
[WIP] poll analysis status

rwblair:
For #115 

Opening a PR so we can test discuss this. This style of polling is an old way of doing things, but its nice in that we don't have to change the server at all.

Right now it is only active for people who are logged in, and it only monitors analyses that they own.

Need to test to see if this causes any issues with memory usage over  the long term.

An alternative would be socketio which has a flask library and uses web sockets. can send updates to the client side when status changes are made.


tyarkoni:
I'd say if it works as-is right now, and it's not obvious that it's going to spring giant memory leaks, it's fine to leave it as-is, assuming there are other things to work on. We can revisit it later if scalability becomes an issue.

adelavega:
I've increased the gunicorn worker timeout to 120s. This should reduce problems for many analyses, but unfortunately (for me), many analyses will still timeout. In the long term I think I need to get SQL (+Flask models etc..) working on Celery...

rwblair:
Still needed on this PR:

-  Only show changes on Fail/Pass (not submitting etc)
-  Grey out generate button immediately after being pressed
-  Selecting a new predictor in the predictor tab should enable the save button

rwblair:
Latest commits should address my previous comments issues.

PR #294, opened 2018-07-18
261 add additional xforms

### LINKED ISSUE #261: Add/update additional transformations
A few additional transformations have been added to pybids that we should add:

These include:

`to_dense`

`scale` also takes the following argument now: `replace_na` with possible values (`before` or `after`)

Here's a full implementation of the transformations and their arguments: https://github.com/INCF/pybids/blob/master/bids/analysis/transformations/compute.py

rwblair:
Fixes #261


adelavega:
Minor comments: The drop down menu is not wide enough to see "orthogonalize"

adelavega:
screenshot from 2018-08-07 19-44-34

Looks a bit funky at the bottom (overlapping)

Issue #285, opened 2018-07-16
Add new fields to variable selection table


adelavega:
New columns: modality, tags, and description (which trails off with ... and tooltip lets you see full description)

adelavega:
Tags should also have the classic 'tag' aesthetic and let you click to filter

adelavega:
This requires the backend to expose these attributes, and the FeatureSchema to add tags.

adelavega:
I think for now just nice descriptions + ExtractorName should be fine

rwblair:
Had been some talk ago about merging the description and extractor columns, I forget if we decided for or against it. Currently in #307 they are seperate columns. Not sure how the two parts in a single column would be styled atm.

adelavega:
I think we would want to stylize it in separate columns (`Extractor Name` and `description`)

PR #295, opened 2018-07-18
Annotate ingested features, add modality/source to Predictors

### LINKED ISSUE #290: Describe / normalize ingested feature in featureschema
Currently the feature schema only applies to ExtractedFeatures. It would be useful to also add descriptions/tags/etc to ingested features such as "RT", and "fmriprep" features.

For features that may have different names across datasets, it would be nice to force them to be the same. E.g. "RT" and "reaction_time", should be "reaction_time" in two datasets.

adelavega:
Fixes #290 and #285


adelavega:
I implemented a function to update objects in the DB against updated schemas.

There are two main limitations:
1. If the name has been changed with a regex transformation (e.g. "rt" to "reaction_time"), the schema may not match the new feature name. I was going to implement optionally matching to the new name, but those can include syntax to insert groups from original regex, so it would be possible to match in many cases. Since descriptions can also include such syntax, its essential to have the original name.
2. A solution to #1 is to avoid updating unless necessary, or adding entries to schema to cover new names, if necessary. 

This all needs to be tested, with unit tests and in a staging version of the database.

adelavega:
Decided to keep track of `original_name` for ingested and extracted features. This solves the previous issues, except for features already in the main DB.

adelavega:
@rwblair Predictors now have pretty descriptions (even those that dont have an Extracted Feature). Remember when you get a chance to update the app to check first the Predictor field for a description, then ExtractedFeature.

adelavega:
@tyarkoni Don't know if you wanna review this, but I added functionality to update the description, name and source of predictors already in DB according to a JSON schema. Seems to work, and will merge soon.

tyarkoni:
Would like to, but I'm pressed for time, so go ahead and merge it and you can fill me in later. Thanks!

adelavega:
Sounds good, no problem!

PR #296, opened 2018-07-25
Updated extractor descriptions

adelavega:
This PR:

1. Fixes a bug in `update_annotations` that prevent matching to `extractor_name`
2. Updated extractor feature schema descriptions.

Left to do:

- [ ] Update GoogleVision Face API feature descriptions. And think of how to re-add features that were inactive (maybe this needs a helper, otherwise it requires re-creating manually. Could force re-create Predictors for the latest ExtractedFeatures. Also, shouldn't dataset_id be a foreign key for EF?)
- [x] Add annotations to transformed features (e.g. `shot_change`). Maybe can use extractor name _trans in schema, and simple check that after transformations. Would make it easy to update after the fact.

### DIFF
--- a/neuroscout/config/feature_schema.json
+++ b/neuroscout/config/feature_schema.json
@@ -3,7 +3,7 @@
     {
       "features": {
         "brightness$": {
-          "description": "Brightness of an image."
+          "description": "Average luminosity of the pixels"
         }
       }
     }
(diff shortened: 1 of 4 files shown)


adelavega:
Remove `_trans` and added `transformed` property to `ExtractedFeatures`. Not going to worry about inactive features, they are actually the rare ones that we dont need to expose yet.

Issue #280, opened 2018-07-12
Pagination deselection


adelavega:
In the run table, it seems that deselecting all runs (top checkbox), only deselects runs in that page, not overall, which is the behavior that would be expected.

Issue #236, opened 2018-05-11
Enable full text search over various Predictor attributes


adelavega:
Currently, only the names of the features are searched. It would be useful to allow full text search over more parameters such as the `extractor_name`, `modality` and maybe even `description`

adelavega:
In relation, the API needs to expose `extractor_name`

Issue #305, opened 2018-08-02
Description and Predictions box are uneditable


adelavega:
For some reason, these two boxes are not editable at all, even after saving the analysis. I suspect its related to the update of the frontend libraries.

Issue #303, opened 2018-07-31
Don't fetch Predictors until final run selection made


adelavega:
Currently, everytime you select/unselect a run, the Predictors are pre-fetched. 

This is unnecessary, and also causes a bug, in which the frontend waits until the Predictors are fetched, then resets the selection. So selection can sometimes get undone if you move too quickly.

PR #307, opened 2018-08-07
make description/predictions editable and delay loading predictors

rwblair:
Fixes for #305 and #303 .

### DIFF
--- a/neuroscout/frontend/src/Builder.tsx
+++ b/neuroscout/frontend/src/Builder.tsx
@@ -46,6 +46,7 @@ const defaultConfig: AnalysisConfig = { smoothing: DEFAULT_SMOOTHING, predictorC
 const initializeStore = (): Store => ({
   activeTab: 'overview',
   predictorsActive: false,
+  predictorsLoad: false,
   transformationsActive: false,
   contrastsActive: false,
   modelingActive: true,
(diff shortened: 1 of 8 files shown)


rwblair:
Going to see if I can't get a loading display going for the predictors table that is sourced from an api call. Also run selection does not appear to affect predictors returned in this branch.

rwblair:
Also includes fixes for #256 and #301 now

adelavega:
On #294 there are two minor aesthetic issues that could be fixed

adelavega:
Lets call it "Run number" instead of "Run ID"

rwblair:
commit 9703d49 fixes #280

rwblair:
predictor search now hits extractor_name, and extractor_description when present. fixes #236?

Set smoothing between 0 to 10 with a default of 5. @adelavega these values ok?

adelavega:
Awesome, will test shortly. Yes those value are OK for now

adelavega:
Cool. All looks good. Minor feedback (which we can address in another PR):

- Make the variable selection list a bit wider. It feels cramped now with the new column.
- Rename ExtractorName to `Source`. And If there is no extractor name, then pull from `Predictor.source`
- When there are no matches, it should say something other than "No data"

rwblair:
I'll handle the variable selection in #284, when we move HRF to its own panel/tab, we might just want to have the predictor table take up the width, and always have the list of selected variables underneath the table except for the widest of screens.

Working on a fix for the Predictor source column and no results copy.

adelavega:
Cool

rwblair:
Ok barring any further comments I won't pile anything else into this pull request.

adelavega:
Sounds god, let me know when its done and I'll test and merge

rwblair:
Ok this is ready for testing.

adelavega:
Looks good, source loads properly. But we can we make that columns Sortable? Also, lets put `Description` last. Even better, if the default sort was source and then feature_name.

adelavega:
Also, it seems full text search does not search the Source column?

rwblair:
Ok f0a1365 fixes search on source and has the table start default sorted by source then name.

adelavega:
Perfect

PR #335, opened 2018-08-24
Add caching using flask-caching

### LINKED ISSUE #319: Add memoization / cache for any slow running queries
Such as `/api/predictors`

adelavega:
Closes #319 
Using `cache` with a custom `key` allows the same arrangement of decorators to remain (and importantly for swagger docs to be generated correctly), while still caching each query uniquely.

Maybe the cache time should be tweaked, though.

@tyarkoni any comments? how long would you cache the queries? default is very short (50s), but I'm sure it could be much longer (especially for things such as datasets, that don't update often)


adelavega:
Cache is now infinite, but gets reset upon feature extraction.

PR #339, opened 2018-08-28
Add statistics at the Predictor level

adelavega:
Closes #232 
Instead of caching a stored values for PredictorRun values. For now, instead, will add dynamic properties to Predictors which use SQLAlchemy to compute these values. A bit costly, but caching should for the most part solve this issue.


adelavega:
Ross commented that in the future maybe having a per run, or per selected runs cache (in db cache) would be useful, but for now this is good!

PR #350, opened 2018-08-31
Run filtering

rwblair:
This is forked from bug-fixesv.2 branch so that should be merged first.

This allows filtering runs by run number, subject, and session simultaneously. 

When you test it pay attention to how select all works, if you change your filters it will reselect everything that matches your filter. I wasn't readily able to find out how to stop that from happening. If it seems like something that should behave differently than it does let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,10 +17,10 @@ In addition, configure where file keys (such as for Google) are located:
     KEY_DIR=/home/me/.keys
     export KEY_DIR
 
-Build the containers and start the services:
+Build the containers and start services using the development configuration:
 
     docker-compose build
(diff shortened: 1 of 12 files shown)


rwblair:
Fixed the changing dataset issue. Now when the run table loads all runs are automatically selected, and all filters are enabled. Added a clear selection/filter button, might need some style changes, let me know if anything comes to mind when you see it.

adelavega:
Looks pretty good to me. Maybe we can worry about styling changes later when we address the accordion etc... It's relatively low priority compared to other tabs I think.

PR #352, opened 2018-09-06
API Change: Dump/load flat lists for predictors and runs

adelavega:
This PR implements a backwards changing API change. Runs and predictors now dump and load with flat lists of ids, rather than lists of objects with "id" as an item. This was original this way due to a marshmallow issue (asymetrical dumping/loading). Looks like this is getting fixed in 3.0, but it's still in beta, and upgrading seems to cause other problems that I rather not deal with now.

Also, I fixed some small schema errors, and the Swagger UI URL for Development mode.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -6,4 +6,4 @@ services:
 
   swagger-ui:
     environment:
-      - "API_URL=https://127.0.0.1/swagger/"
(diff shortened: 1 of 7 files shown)


adelavega:
@rwblair just want to make you aware of this change. let me know if it causes problems for you.

Issue #320, opened 2018-08-20
Add save/next button to bottom right


adelavega:

adelavega:
Closing in favor of #380

PR #360, opened 2018-09-11
Fix bug with auto selected run IDs.

### LINKED ISSUE #359: Bug: No predictors are fetched, if name is given after run specification
If I do the following:
New Analysis -> Select Dataset -> Select runs

I notice that the "Next" button is not yet available. Is this on purpose? I'm OK with this, but perhaps the button should be there, but greyed out.

If I then fill the name out, the button appears. But now going to the next tab, the predictors are not fetched.

If I do this in the normal order ("name", then dataset/run selection), this problem doesn't occur.

rwblair:
fixes #359

The code added that automatically selects runIds when a dataset is selected happens is a promise and was getting executed after the code that checked to see if new predictors should be loaded. Went ahead and set the flags to load predictors and allow access to the predictors tab.

### DIFF
--- a/neuroscout/frontend/src/Builder.tsx
+++ b/neuroscout/frontend/src/Builder.tsx
@@ -544,8 +544,12 @@ export default class AnalysisBuilder extends React.Component<BuilderProps, Store
             message.success(`Fetched ${data.length} runs associated with the selected dataset`);
             let availTasks = getTasks(data);
             if (availTasks.length === 1) {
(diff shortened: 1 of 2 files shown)


adelavega:
Going ahead and merging, but let's circle back to the Button issue (i.e having consistent UI) when addressing #320 
Basically, the "Next" button should always be in the same place for all tabs (I don't care if its left or right though)

PR #368, opened 2018-09-12
Update pybids to 0.7

### LINKED ISSUE #367: Test pybids=0.7
Test and refactor to use pybids==0.7

adelavega:
Closes #367 

TODO:
 - Since `exclude` is being deprecated we need a way to specific which event files to look for in `fitlins` (see: https://github.com/neuroscout/neuroscout-cli/issues/52). This may require changing the file names for bundle files.

### DIFF
--- a/celery_worker/tasks.py
+++ b/celery_worker/tasks.py
@@ -6,7 +6,7 @@ import json
 from pathlib import Path
 from tempfile import mkdtemp
 from bids.analysis import Analysis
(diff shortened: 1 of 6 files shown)


adelavega:
Since `desc` is not coming until derivatives stabalizes, we can deal with the naming when that comes out. In the meantime, I guess we will pin fitlins to the last version using pybids < 0.7

Issue #343, opened 2018-08-30
In dataset list, link to original resources


tyarkoni:
I.e., add a column for URL of the resource to the table, so we don't have to worry about extended description or explanation

adelavega:
Also, fix description for task. If only one task-- repeat

PR #372, opened 2018-09-14
Add dataset resources, and task description

adelavega:
Addresses #343 
Instead of repeating dataset description, added a hand annotation for task "summary"


adelavega:
Hey @rwblair, could you update the frontend to show `summary` for Task "description". And add a small column on the right for each Dataset that is a link to the `url` property? Feel free to push to this PR.

adelavega:
Also note that the database needs to manually updated to have those values (although upon fresh ingestion, they would be there)

rwblair:
The URL column is a bit cramped with all the other fields on the dataset table. Push with it in place so you can see it as is. Width is hard set and should probably be increased.

Another option might be to have the name of the dataset be a link to the URL.

Issue #378, opened 2018-09-18
Reports routes for review tab


adelavega:
The goal will be the show modulate reports on the review tab.
To do this, we have to decouple `/analysis/compile` and `/analysis/reports`
Instead, create many parallel routes on `/analysis/reports` that execute in parallel in the background (and are cahced).

Example sub-routes on `/api/analysis/{hash_id}/{}`
- Transformations
- Events
- Diagnostics
- Description
- Design Matrix

At first, could return dummy results for these, but at least define API for @rwblair to begin building Review tab #328

adelavega:
@effigies, @tyarkoni and I discussed showing users interact reports of their BIDS model design prior to analysis execution. Since reports will be generated on designs prior and during execution in `fitlins`, it naturally suggests reports should be modularized outside of fitlins execution per se. 

Would it make sense to put these reports into `bids.analysis`, which then `fitlins` and `neuroscout` can both use independently?

effigies:
I'm okay with that, though pybids is starting to get pretty heavy...

tyarkoni:
In a sense, pybids already supports this; the main thing we'll want to show users is the design matrix, and fitlins is already getting that via `get_design_matrix` anyway. The only thing pybids doesn't do for you (and probably shouldn't) is visualize the design matrix as an image. But once you have the design matrix as a `DataFrame`, you can just standardize the columns and show the whole thing as a heatmap, and that would be perfectly adequate for our purposes.

That said, I think rather than adding parallel calls from NeuroScout to pybids, it would probably be easier if fitlins just wrote an image of the design matrix out to disk, and NeuroScout could just grab that and display it.

adelavega:
Well, I rather conceptualize it as decoupling some of fitlins reports generation from the actual execution / nipype, so that the call the generate these reports is as light as possible. 
That is, I could just import from fitlins the functions I need (and maybe theres a higher level call that generates several reports at once)

I have taken a look at fitlins code, but maybe that's already the case.

effigies:
The nipype interfaces are very thin wrappers on regular functions.

Visualization functions are here: https://github.com/poldracklab/fitlins/tree/master/fitlins/viz

adelavega:
Perfect. Maybe I should looked first.... :)

adelavega:
[...]

Issue #377, opened 2018-09-18
Parametarize `/api/compile`


adelavega:
Currently, compile only runs on the first run.

We should also allow users to select on how many users to perform compilation on by adding parameters this route.

adelavega:
Closed w #382

PR #382, opened 2018-09-19
Generate analysis reports

adelavega:
WIP to address #378  and #377 

Reports can be requested by `POST` to `/api/analyses/{analysis_id}/report` with query string list of `run_id`. This creates a new background task, and records Report in table.

a `GET` to `/api/analyses/{analysis_id}/report` (again with matching `run_id`), will look for the newest Report in the table that is fresh (i.e. newer than `modified_at` for analysis). If no fresh Report can be found a `404` is thrown. Otherwise, the `Report` object is serialized.

The serialized `Report` object consists of links to resources created in the report.
@rwblair give it a shot. Currently, it only returns a "dummy" report, but you can start implementing the logic using these routes. 

Still need to:

- [x] Implement creating actual reports / resources
- [x] Host these resources 
- [x] Generate correct links to these resources (these could also be "links", to be truly RESTful)
- [x] Add tests


adelavega:
Also, regarding compilation, given this model, to be consistent we could do the following

To request compilation do a `POST` to `/api/analyses/{analysis_id}/compile`
To check status, do a `GET` to this same route, rather than a `status` route.
I like this because its more concise and uses the HTTP verbs correctly. 
I also think this doesn't need to be be parametarized (like reports), because all runs should just be checked. It may take 1 minute longer, but it's for the best. The longest part is writing out the bundle which needs to be done regardless. 

Thoughts @rwblair? @tyarkoni

rwblair:
Sounds good to me.

adelavega:
Allright this PR is almost done... except for producing any useful reports, because those are sort of embedded within fitlins

Issue #284, opened 2018-07-16
Move HRF selection to separate tab


adelavega:
And focus the variable selection tab on just selection.

HRF selection tab does not need to display "description" and other such things.

Issue #376, opened 2018-09-18
HRF as transformation


adelavega:
The `hrf_variables` field is (probably) going away. Either way, going forward it will be most useful to define HRF convolution as a transformation.

The plan is to add a tab *after* transformations (so that all possible variable names are available).

At first, let's just insert the transformation at the end of the list (without showing in the main `transformation` tab).

In the future, we can allow users to select where to insert, and show a complete list of transformation.

PR #383, opened 2018-09-19
Move HRF selection to its own tab.

rwblair:
For #284


rwblair:
Also does #376

adelavega:
Looks like HRF table is still a bit short? Like only 3 rows for me? It's live on alpha if you want to check it.

adelavega:
oh im really dumb, its because I only selected a few predictors...

PR #384, opened 2018-09-30
Review/Reports Tab.

rwblair:
This branch has the changes from #382 merged into it.

I've added an alias in the nginx config to allow downloading from the file-data/analyses directory, Will want to be sure to change that back before merging this branch/once we come up with an alternative way to serve private bundles.

Download link on reports tab needs something done to it, but I'm not sure what.

Still working on forcing a linear step through of all the tabs.


rwblair:
Also this PR is manually building the download link for the full bundle. We might want to add the bundle download url in a response from /reports or /compile.

adelavega:
Hey Ross, overall it looks good

My main comment is that the Report tab should be second to last, and should POST to `/report` to request a report be generated. The `/compile` route is the final step, and does not depend on the `/report` in any way. Here are my proposed changes:

1) Reorder and rename tabs: `Review` and then `Submit`
The Review tab should contain what is currently in `Review` + the results from `/report`. By default, navigating to this tab should request a report for the first run in the analysis, but there should be an interface to request a review for a different run (e.g. a drop down that allows you to select a run, or maybe multiple runs.

Maybe one way to organize this is to either combine the two accordions which are currently spread out over the two tabs, or to put them one after the other, and separated. I would suggest the results from "Report" could go higher, since they are more important, and a "Summary" section could be below, with the accordions.

2) Submit tab.
I can work on this, but I think the `Submit` tab will be less informational, and rather ask you to confirm you agree the our User Agreement (TBD), and have some info about how to run your analysis (and display the status)

Bugs / Minor comments:
1) Whenever I clicked "Save" the content of that tab disspeared
2) Display subjects and runs as comma separated, not bullets (to use horizontal space). 
3) Move the "variables" section to be higher in Review (after "Inputs")

adelavega:
Another bug: when I click `Generate` the Analysis I/D gets cleared and becomes `Analysis ID: n/a`

adelavega:
Like I said on Slack, it was the `hrf` transformation missing that caused the report to fail (although we should probably find a better way to deal with failures, and error codes, since they are quite possible)

Another issue, the image of the design matrix that was inserted had an incorrect link, and didn't display correctly e.g.:
`https://alpha.neuroscout.orgalpha.neuroscout.org/reports/M3ZRA/sub-19_task-MerlinMovie_design_matrix_plot.png`

Note the repeated domain name

rwblair:
> Once I clicked generate, it did not refresh automatically once it was done. I refreshed manually and it looked good.

I think if you wait long enough it will, might take 2 cycles on the slow polling. I'll add a special case when you submit to look for updates on that ID quicker. Ideally would replace this with websockets some day.

rwblair:
@adelavega Ok this should address all of the issues in the previous review.

adelavega:
Looks great! Very solid and much more mature feeling.
I have a few comments, some of which we can get to later

More urgent
- I made some aesthetic tweaks. Let me know what you think
- It seems the Report tab is enabled even if you haven't gone through all tabs. Arguably we don't want to allow users to skip ahead to this tab. 

Urgent but can do in a separate PR:
- We should validate fields one tab at a time. That is, users should have to put a name in before advancing to the next page. This would allow us to use that form validation aesthetic where the field turns red. I think we should avoid the pop up errors, they  are a bit jarring (altough I like the pop up auto save notification)
- The previous point actually leads to a bug, or bad UX. If you dont give it a name, then get to reports, it will error out while still showing the loading icon. 
- You can also advance past Predictors without selecting any, which should not be allowed, and can lead to errors on subsequent tabs

Minor
- It's been bugging me how the top is wider than the form. We should align those. Especially since it wouldn't be a bad idea to have a Nav Bar in the future. 
- Everything seems a tad too wide sometimes, but maybe it's just me. 
- Analyses default as private? I think we should default to public.

rwblair:
Form validation should be done in this PR since for the overview validation was working before the PR. Will do the red highlight, for name field. Not sure how we might indicate required to advance on the predictors tab, might still be able to highlight that whole table red in antd, though its literally the only thing you can click on in the screen.

Tweaks made are good. Differentiating the add xform/contrast button is real good, esp since I didn't change the colors on the prev/next buttons, should I still make those green?

Narrowing the header is no problem and makes sense, and can narrow everything by the standard column size or two.

adelavega:
I think we're good on the button color. I looked and noticed ant doesn't have green buttons. For some of consistency, lets stick with the built ins for now. 

I think for predictor page validation page, its simple. Just disable the button if no predictors are selected. 
I think it's pretty obvious what is required, especially since theres only one thing to do on that page.

Re: narrowing. I think the main content is only a *hair* too wide, so no need to overdoit. Especially with how most people have pretty wide screens these days. It mostly looked too wide on my external monitor.

Thanks! Overall, it's coming together real nice, and I'm glad I can understand the code enough to tweak the aesthetics. I'm going to do a bit of digging into what else ants has to offer.

rwblair:
One more time with feeling. Ready for another look at.

rwblair:
Ok fixed the styling and added validation on the xform and contrast adders. transform I think its clear you should select a predictor, on the contrast editor I added in the red start for the name and weights like the analysis name in the overview, the red box with messages on what should be added no longer appears. Not sure if its too subtle. Be interesting to see how someone who has never used the site would interact with it. (guess I got a lab full of people I could do testing on...)

adelavega:
Cool, thanks, I'll check it out. Yeah I think some testing would be good to do once we get more solid. I think most ppl would figure it out reasonably well, I'm just a fan of maximizing how natural a UI is. Reduces friction.

adelavega:
Awesome, looks good to me! 
I do have one note for later: maybe it would be better to let ppl click "OK" and have form validation errors pop up, rather than disable the OK button. I actually forgot to put the name for a contrast and it took me a second to figure out why the button was still disabled. But, that's nitpicking for later!

adelavega:
Okay, now the frontend tests are failing, but once you fix it, go ahead and merge!

rwblair:
Front end tests are passing, merging.

Issue #355, opened 2018-09-08
Update frontend editability rules


adelavega:
With #354, I added a `locked` field, which can be enabled by admins. This will be rare, but will disallow any edits regardless of `status`.

Aside from when `locked` is true, the fields `description` and `private` are always editable now. 

Let's make sure the frontend respects these. Although, let's think about how this plays with the view only mode for post-compiled analyses.

PR #387, opened 2018-10-16
[WIP] Single sign on with Google

### LINKED ISSUE #163: Add single sign on
Starting with Google, support single sign on (based on email identity). Hopefully this is easy to do with JWTs.

adelavega:
Closes #163 by implementing OAuth2 with Google.

I designed this in a way to change as little as possible from our current API so its a slightly hacky.

To do a single sign on with Google, you embed a Google button. Test page is at `localhost/google`
This returns a JWT in the console that can be passed back to the server.

The JWT can be passed back to the server using the `/api/auth` route like so:
`{"email": "GOOGLE", "password": "JWT_HERE"}`
Note email must be `GOOGLE`.

If the user is found (same google ID), a regular JWT will be returned which can be used for regular purposes. If a user with the same email but a different user ID is found, that user will be updated to be associated with this google account, and a JWT will be returned.

If no user is found, a new user is created, and is auto-confirmed if Google has confirmed their email.

@rwblair Let me know what you think. Although slightly hacky, it seems like this is the least amount of work to get this to work. Alternatively, I could use a different package (or fork it), which would allow me to change the JWT verification function to accept Google JWTs.


adelavega:
Tests are failing because of pybids related issues, and should probably fix themselves once I merge the other branch.

rwblair:
Looks straightforward, I say we stick with it.

adelavega:
Cool, just in case it wasn't clear, I'm taking the Google JWT, verifying it, checking its claims, and then issuing a regular Neuroscout JWT for further use.

adelavega:
If we wanted to incorporate SSO from ORCID: https://members.orcid.org/api/integrate/orcid-sign-in

adelavega:
Let's use badges for user pictures (from SSO), defaulting to a generic user icon for regular users: https://ant.design/components/badge/

adelavega:
Sorry I meant "Avatar"

rwblair:
- [ ] Remove console.logs and set the short poll timing back to normal when done testing.

rwblair:
So far this will not reload the google avatar url and display email from a cold load with a saved jwt.

rwblair:
Bug: login -> logout -> login new acct results in the same JWT.

rwblair:
Fixed above bug, and got the error messages off for logged out user.

Still working on only allowing the analysis owner update name/desc/pub-priv on a published dataset.

rwblair:
@adelavega ok I think I've got this close to where I want it. When ever you get a chance can you look at it? Particularly when a google user's token expires, see if there's anything we want to change about how that looks.

adelavega:
This PR will also close #391 #388 #355

rwblair:
Reason why these are failing is from some missing typescript definition. I manually updated an entry in my node_modules/antd to get it to work. Not sure when this changed. Still looking for a workaround.

adelavega:
Also closes #405

PR #390, opened 2018-10-25
Run selection accordion

### LINKED ISSUE #351: Add collapsible sections to UI
To streamline the UI, and to prevent the page from being too long, implementing accordions to hide selected sections would be useful. Ideally the page length does not require scrolling so that the "next" "forward' buttons can be seen (although I'm not sure we need to be super strict about this either).

I've mocked up the Bootstrap Accordion UI for select dataset/task/runs here: https://codepen.io/anon/pen/VGzyQa

Note that since there is only one task, it is already preselected, and that is reflected in the title of the task section. Also, by default all runs are selected. Since most users will be OK with these default settings, it streamlines this section. When each new section is opened, all other sections close.

Ideally, when a dataset is chosen and the card is closed, the dataset name would be inserted into card: "Dataset - {dataset_name}"

adelavega:
To make the UI fit in a single page, and not overwhelm users with information, I implemented a collapsible accordion for task and run selection.

Closes #351 
This was branched off #384 

@rwblair Can you help me change the title of the run panel, so that it reflects the number of runs selected?

Maybe if all runs are selected (default) it says:
`Runs: All selected`
then if less than all are selected it says:
`Runs: 15/17 selected`

It could also say `17/17` for the first option.


adelavega:
Looks nice. but it looks like the function is not evaluating (same problem I had, and why I passed it off to you!) 
image

rwblair:
Just realized after I pushed to use that style of string formatting I have to switch the ' to a ` I think. commit I just made uses old way of making strings.

rwblair:
Now I've noticed a bug in SherlockMerlin sorting by subject deselects all runs. In Life sorting by subject selects all runs.

adelavega:
That's odd. Can you also change the "Select task" to say "Task: {name} selected" If no task is selected (when more than one task) it can say just "Task: Please select"

rwblair:
Ok task name should be listed when selected. Lets hold off on merging this until Monday, I want to think about how these variables are being handled a little bit longer.

adelavega:
Looks good to me! Feel free to merge if you are ready

PR #392, opened 2018-10-30
Menu/tab UI overhaul

adelavega:
This PR is a relatively big aesthetic change to the UI.
The main change is the use of a Menu bar at the top, and flipping the Builder tabs to be vertical.

The main things left to do for now:

- [x] Add ability to disable "Previous" button for first Builder tab
- [x] Plug in "ID" tag at the bottom of the buider
- [x] if you go the URL without clicking through the menu, the appropriate menu icon should be highlighted (not as important, but a bit confusing because otherwise it stays highlighted)

Later on we can fill the "My Profile" and "Help" tabs with content.

@rwblair can you take care of those things, and take a good look and give me some feedback? I'm not 100% sure the routing is correctly implemented

### DIFF
--- a/neuroscout/frontend/src/App.tsx
+++ b/neuroscout/frontend/src/App.tsx
@@ -401,9 +401,12 @@ class App extends Reflux.Component<any, {}, AppState> {
                 >
                   <Menu.Item key="home">
                     <span className="nav-text" style={{fontSize: 20}}>Neuroscout</span></Menu.Item>
-                    
+
                   {this.state.auth.loggedIn ?
-                    <Menu.SubMenu style={{float: 'right'}} title={<Avatar shape="circle" icon="user" />}>
+                    <Menu.SubMenu
+                      style={{float: 'right'}}
+                      title={<Avatar shape="circle" icon="user" />}
+                    >
                        <Menu.ItemGroup title={`${email}`}>
                          <Menu.Item
                           key="mine"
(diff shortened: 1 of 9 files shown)


adelavega:
Hey I found a bug related to redux. 
`TypeError: auth_actions_1.authActions.signup is not a function`
When signing up for a new user.

rwblair:
KK I'll fix that. Notes from run through with Franklin:

- Some sort of explanatory text in reports about what the user is looking at.
- replace the phrase 'like this' with 'with the following command' in the successful compilation view.
- might be cool to have a 'choose output directory' on the last page, and insert that path into the example docker command.
- contrast weights can't be negative again.
- More text/info on a failed report image generation failure to let the user know what to do next.
- Some lists in the model.json aren't ordered alphanumerically. (for a second we thought maybe all the subjects hadn't been included)

adelavega:
Awesome, that sounds good. I think some of the doc related suggestions can be saved for later. I think adding tooltips w/ helpful tips/explanations throughout the UI will be very important. Same for execution.

rwblair:
confirmed 0f9edfc fixes user registration when email sending is disabled in my local setup, and the user set to confirmed in the database. Appears the 'CONFIRM_USERS' setting is not being used.

adelavega:
YesI think that's an outdated key now that I look at it. I think its' because its just set to confirm regardless (well unless on SSO on the other branch)

Issue #233, opened 2018-05-11
Additional FAQ questions


tyarkoni:
Basic user-facing documentation.

adelavega:
- How do I get started?
See the Neuroscout tutorial here
- Is this service free to use?
Yes!
- Are there any restrictions on analyses I create on Neuroscout?
Yes, by using Neuroscout, you agree that once you have finalized and "compiled" an analysis, the analysis can no longer be deleted from our system. If you wish to 'edit' an Analysis, you may clone the analysis, and make any desired changed on the forked version. Although analyses are by default not searchable by other users, any user with the private analysis ID may view your analysis.
- I plan to publish results I've obtained using Neuroscout, how do I cite Neuroscout?
Please cite (zenodo link for now?), and be sure to include all relevant analysis IDs. To maximize transparency and reproducibility, we ask that you do not fork analyses generated by Neuroscout offline, and instead use the "clone" function to modify and generate a new derivative analysis. 
- I have a naturalistic study I'd like to share on Neuroscout, how do I do so?
Due to the financial cost of extracting features from multi-modal stimuli using external APIs, the datasets we support is a manually curated set. However, we are actively expanding the datasets supported, so please contact us.
To facilitate the process (and to make your dataset otherwise publically accessible), we recommend uploading your dataset to OpenNeuro.
- I want to make changes to an analysis I already ran, but it is locked. How can I edit it?
Once an analysis has been run, it is permanently locked and archived for provenance. You may "clone" your analysis, and make changes to this new copy of your analysis.
- I want to share an analysis I already ran with the public, but it is locked. 
The two aspects of analyses that are always editable are the name, description, and public/private status. So go ahead and make your analysis public!

tyarkoni:
Good start; I made some edits. Some other likely FAQs:

* Am I restricted to mass univariate GLMs, or can I use NeuroScout to specify other kinds of analyses?

* How do I cite the tools I'm using? [We should probably auto-generate a reference list for every analysis, to be displayed on the status page. This would at minimum include a NeuroScout citation, any citations to the dataset, and the pliers citation if one or more pliers features were used.]

adelavega:
Aesthetically, we might want to revise this to a TOC w/ links, or the accordions being expanded by default.

PR #394, opened 2018-11-01
FAQ + Misc changes

adelavega:
This is a PR to take care of a few miscellaneous improvements including:

- [x] Misc wording changes
- [x] Adding FAQ to Help submenu
- [x] Add "selected predictors" subheading above the tags next to the table (a bit confusing otherwise)
- [x] Make new FAQ page (with content from #233) and a placeholder User's Guide page (or we can link to an external place?)
- [x] Move public/private button to `Run` tab
- [x] Remove F tests for now (I mean just remove the radio button, we'll add it back later, but it requires more weights we don't currently have a way to set)
- [x] Fix new analysis bug #395
- [x] Maybe make everything a bit wider? With the tabs being vertical, we have a few issues where width becomes an issue (e.g. "Description" field). I would maybe make the Columns a hair wider across the board (especially at the 2nd smallest width, where things get real cramped), and tweak the width in the tables to fix any issues. We could also use `...` and tooltips for things that are too long...


adelavega:
hey @rwblair can you maybe also implement the "My Analyses" link in the Profile dropdown. It would simply be a page with the `Your Saved Analyses` display, without the content above it. Soon, I'll fill in the content above it for

PR #397, opened 2018-11-08
FIX: Display static content

adelavega:
A minor fix to display static content correctly. They need to be in build/static to be properly served.


adelavega:
@rwblair look good to you?

PR #400, opened 2018-11-14
Update BIDS StatsModel produced by frontend

adelavega:
So far, I've gone through and CamelCase'd a bunch of stuff, and changed some names (e.g. `Steps`).

TBH, I haven't put much thought into this aside from go off the compiler's errors, but so far it seems to be working ok. 

Please take a look @rwblair

### DIFF
--- a/neuroscout/frontend/src/App.css
+++ b/neuroscout/frontend/src/App.css
@@ -79,7 +79,7 @@ pre {
 }
 
 .splashLogo {
-  display: block;
+  display: step;
   margin-left: auto;
   margin-right: auto;
   width: 30%;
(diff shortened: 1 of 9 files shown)


adelavega:
It wasn't always clear what was an internal variable and what was a variable for the model (which should be CamelCased). It's possible I missed some parametrs, or camelcased some stuff that I shouldn't have (although I suppose that should't be too much of a problem as long as its consistent)

adelavega:
Looks like the Contrast builder is having some issues probably because some names dont match.

adelavega:
Well, it builds models! and the review pane looks good so far. I'm going to push it to the alpha server so I can test the backend compatibility.

rwblair:
Changes look good and straight forward to me, camel casing internal variables is fine by me. Seems like the closer we can keep them and what will end up being used in the spec the less confusion there will be.

PR #409, opened 2018-11-25
Fix large model handling

adelavega:
On models above 50+ predictors, submitting sometimes fails due to redis. It may be that too much data is being transfered over that protocol.

For now, I'm going to add a `submitted_at` date, and catch hanging submissions. They will return a `FAILED` status and show a reason (model being too large)

Need to test this on alpha, and determine how to fix redis. Also need to see if in production version gunicorn workers timeout or not.

I'm also sneaking in some pybids compatibility upgrades


adelavega:
Tested on alpha. Looks to be working. Medium analysis was submitted, while very large was not, but gave proper error.

PR #412, opened 2018-12-04
Add a replace n/a with 0 transform before compilation.

### LINKED ISSUE #329: Add "replace na" a transformation (final step)
By default, NAs should be replaced by 0s after transformations. This should be a checkbox in the 'config' tab.

Or perhaps this should be a transformation, so that it is more explicit in the BIDS model.

rwblair:
fixes #329. Looks to see if a transformation exists that has all the predictors in it and replaces 'n/a' with 0. If it doesn't exist that transformation is added on submission for compilation.

### DIFF
--- a/neuroscout/frontend/package.json
+++ b/neuroscout/frontend/package.json
@@ -24,6 +24,7 @@
     "@types/enzyme": "^3.1.10",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
+    "jest-fetch-mock": "^2.0.1",
     "react-scripts-ts": "^2.3.2",
     "tsutils": "^2.26.2",
     "typescript": "^3"
--- a/neuroscout/frontend/src/AnalysisList.tsx
+++ b/neuroscout/frontend/src/AnalysisList.tsx
@@ -46,7 +46,14 @@ const AnalysisList = (props: AnalysisListProps) => {
       title: 'Status',
       dataIndex: 'status',
       render: (text, record) => <Status status={record.status} />,
-      sorter: (a, b) => a.status.localeCompare(b.status)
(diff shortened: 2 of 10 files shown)


tyarkoni:
Replacing n/a with 0 has the potential to dramatically change the design matrix in ways the user won't be expecting. Why can't it be handled with a `DropNA` transformation?

tyarkoni:
I think I'd prefer to fail gracefully if there are any `NaN` values in the design matrix... maybe this can be part of a check before compilation, and if `NaN`s are found, the user gets an error message telling them to explicitly call `DropNA`.

tyarkoni:
I guess a related question is, where are `NaN` values being introduced? I assume they're coming from features we've extracted and are storing in the DB... if there are some places where null values are common, maybe the best solution is for us to handle that in preprocessing. I vaguely recall discussing this in the past, but don't remember the conclusion of those discussions...

adelavega:
Yes, actually this is the solution we came up with a while ago (although I can't remember the details either...), but looking at it now I had the same concerns you raise.

There are many places `n/a`s could be found. Two examples would be `FramewiseDisplacement` and another is `subtlexusfrequency_Lg10WF`.

I think for some it makes sense to drop in advance from the db, such as `FramewiseDisplacement` (because I can't imagine why anyone would want to fill those with a value), but for others such as frequency its up to the user to deal with that correctly. 
Unless we waned to be more heavy handed and do something like pre-scale + replace n/a with zero for all runs in the db. This would make model building easier but would be quite opinionated and more work for us to mantain.

I agree the best solution is to catch this at the report stage and offer a suggestion. And maybe for *some* variables preprocessing would make sense. This can come as part of a larger PR to give more useful error messages / warnings at the report stage. 

Closing this for now. Sorry for the confusion @rwblair !

PR #413, opened 2018-12-04
[WIP] Allow transformation editing

rwblair:
Still need to write tests for this, will wait for #412 to be merged since it updates frontend tests.

We might also be running into blue button overload on the transformation edit page.

Finally when editing a transformaiton, we can go to the next tab without saving the the changes. Either need to force this save, or grey out the next button until saved @adelavega opinions on which you'd prefer?


rwblair:
changes I am making for #415 supersede this.

PR #433, opened 2018-12-18
Allow reports to be generated multiple times per session if analysis is updated.

### LINKED ISSUE #426: refresh report when design changes
Currently if you go back and change something, the report will not change when you advance back to it

rwblair:
always set 'postReports' true when analysis is updated and saved. Always save analysis when going to report tab.

fixes #426


adelavega:
Maybe we can group #438 into this one. Hopefully its easy.

rwblair:
Loading an analysis was triggering a call to the buildModel function which needed all the available predictors to work as expected. Last commit prevents this from happening on non Draft analyses since the model coming from the server is final and shouldn't change locally.

adelavega:
Tested on alpha. LGTM

Issue #415, opened 2018-12-05
Represent Transformations/Contrasts as draggable table/list/tabs


adelavega:
Tal suggested to represent transformations/contrasts as a list of draggable items (with a few actions attached, such as "delete" "edit"). These would be vertically oriented, sort of like a set of nested tabs (within the main tab navigation).

The key here is that clicking on a transformation (to edit), would not able the list of transformations go away. Ideally the edit would be done either on the right or in a modal.

I worry about ensuring that we visually differentiate this enough from the main UI components, so lets explore using tables, lists and tabs to do this. 

According to this: https://github.com/react-component/tabs/pull/116 it should be possible to make these draggable.

adelavega:
Relatedly:  "Generate Auto-Contrasts" button would populate a list of contrasts. However, if these get deleted, we can't use the "auto_contrasts" directive in the model. Thus, might want to fill them in explicitly in the model. But for debugging purposes, can we keep track that the generate auto contrasts button was clicked?

adelavega:
Closed with #439

Issue #96, opened 2017-07-11
Allow user to edit existing transformation


skept:
@tyarkoni Since it's not too much work for the user to just delete a transformation and add a new one I'm assuming the edit functionality is lower priority?

tyarkoni:
Sounds reasonable, definitely not a high priority for now.

tyarkoni:
Oh, but if editing is not possible, then we want to make sure the user can easily re-order transformations in the list, since the order in which they're applied can be pretty important.

skept:
Good point, I can see how reordering is probably even more important than editing. I added a separate issue: #98

adelavega:
Closed with #439

PR #439, opened 2018-12-20
[WIP] Drag, drop, and edit xforms and contrasts

rwblair:
addresses #415 and #96.

Contrasts view still needs work on how we want to visualize existing contrasts.

### DIFF
--- a/neuroscout/frontend/src/Transformations.tsx
+++ b/neuroscout/frontend/src/Transformations.tsx
@@ -56,8 +56,6 @@ function weightsRequired(name) {
 }
 
 export function validateXform(xform: Transformation) {
-  // tslint:disable-next-line:no-console
-  console.log(xform);
   let errors: string[] = [];
   if (!xform.Name) {
     errors.push('Please select a transformation');
@@ -73,6 +71,17 @@ export function validateXform(xform: Transformation) {
   if ((xform.Name === 'Orthogonalize') && xform.Input !== undefined) {
     if (xform.Other === undefined || xform.Other.length < 1) {
       errors.push('Must orthoganalize with respect to at least one predictor');
+    }
+  }
+  if (xform.Name === 'Replace') {
+    let keys = Object.keys(xform.Replace);
+    let predecessor;
+    let replacement;
+    if (keys.length > 0) {
+      predecessor = keys[0];
+    }
+    if (predecessor === undefined || predecessor.length < 1) {
+      errors.push('Enter value to be replaced.');
     }
   }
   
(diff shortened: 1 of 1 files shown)


adelavega:
For both Contrast and Transformation, when navigating away (either to another transformation or to another tab), save that transformation + validate. In other words, copy + show validation errors, and don't allow navigation unless errors are cleared. This includes for contrast name (allow clicking OK but show error if there is no name or if name is not unique).

rwblair:
Backup alternative would be to notify of unsaved changes on navigating away from unsaved xform/contrast.

rwblair:
For future reference, was seeing if its possible to apply only part of a schema to automate the validation of xforms onSave, but it looks like the BidsModelSchema would need a substantial rewrite to make this happen:
https://github.com/epoberezkin/ajv/issues/211

As of now we have hand coded all the type info for the transforms and their arguments, so for now will just be looking at the required field in each transforms schema for validation.

adelavega:
hey, any updates on this?

rwblair:
@adelavega take a look at this and see what you think.

rwblair:
Ok addressed the original three bugs. Removed the search when there are less than 20 predictors to choose from. Got the weights to be closer together.

Haven't figured out how to get that label on the same line as the contrasttype.

For autocontrasts, what do we want functionality to be if inputs are added or removed after auto contrasts has been clicked? Should we automatically add new auto contrasts if they go back to the overview tab and add more inputs?

adelavega:
Thanks!

Hmm, good question. I would say if its a check box, then yes. We could also make it a button which indicates its a one time action. I think the first option is probably best... that is auto adding if more predictors are added.

rwblair:
Sorry missed you comment before last commit. I'll switch it back to be a checkbox and see what we can do about adding any new predictors to the the autocontrasts list when selected.

adelavega:
I'll ask Tal his opinion this afternoon.

Relatedly, it seems that contrasts is adding weird stuff to the JSON model.

See: https://alpha.neuroscout.org/builder/wyJWA

This is a portion of the JSON Model
```
"Steps": [
    {
      "AutoContrasts": true,
      "Contrasts": [
        {
          "ConditionList": [
            "dark",
            "daylight"
          ],
          "ContrastType": "T",
          "Name": "day_vs_dark",
          "Weights": [
            1,
            -1
          ],
          "selectedPredictors": [
            {
              "description": "Clarifai image recognition label: dark",
              "extracted_feature": {
                "created_at": "2018-05-09 22:42:17.872489",
                "description": "Clarifai image recognition label: dark",
                "extractor_name": "ClarifaiAPIImageExtractor",
                "id": 13647,
                "modality": null
              },
              "id": 12787,
              "max": 0.9882182,
              "mean": 0.3242534958889997,
              "min": 0.0058510695,
              "name": "dark",
              "num_na": 0,
              "source": "ClarifaiAPIImageExtractor",
              "stddev": 0.25294245066492155
            },
            {
              "description": "Clarifai image recognition label: daylight",
              "extracted_feature": {
                "created_at": "2018-05-09 22:42:17.881578",
                "description": "Clarifai image recognition label: daylight",
                "extractor_name": "ClarifaiAPIImageExtractor",
                "id": 13648,
                "modality": null
              },
              "id": 12788,
              "max": 0.98870385,
              "mean": 0.8017381956495143,
              "min": 0.021070436,
              "name": "daylight",
              "num_na": 0,
              "source": "ClarifaiAPIImageExtractor",
              "stddev": 0.14537676107116657
            }
          ]
        }
      ],
      "Level": "Run",
```

PR #450, opened 2019-01-08
Re-design static front page

adelavega:
Aesthetic re-design of front end. Aims to keep it simple (aka somewhat generic/familar), yet decent looking.

Feedback on text, or if something else entirely is desired. 

Could buy icons if they are liked, although theres a few more options as well with the same style.


adelavega:
neuroscout_fp

Here's a preview

rwblair:
Not a fan of the horizontal line at 100% width below the "browse public analyses" button. 
Be nice if OpenNeuro, DataLad, and Neurovault in the copy were links.

I like the three columns of descriptive text with icons/diagrams personally.

adelavega:
Agreed! Maybe if I put it inside a Column that'll fix that. Thanks!

adelavega:
Okay, I made those changes, and also added a new page for browsing "private" analyses since the front page does not serve as that any more. It's live on alpha if you want to check i tout. There may be a more efficient way to do the private / public lists, without two separate files as I did it...

PR #458, opened 2019-01-18
Update README documentation

adelavega:
The documentation was out of date, so I've updated it here to make it a bit more comprehensive. 

@Shotgunosine take a look please and let me know what else I can do to make it easier to set up. 

Edit: Also addressing a few ingestion issues that came up, including:

- Grabbit/pybids deps
- Skip stimuli ingestion if there are none
- Update manage.py/add_task parameters
- Fix regex for finding non numeric values (for automatic computations)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,9 @@
 [![Build Status](https://travis-ci.com/PsychoinformaticsLab/neuroscout.svg?token=mytABRBRnBitJJpBpMxh&branch=master)](https://travis-ci.com/PsychoinformaticsLab/neuroscout)
 [![codecov](https://codecov.io/gh/PsychoinformaticsLab/neuroscout/branch/master/graph/badge.svg)](https://codecov.io/gh/PsychoinformaticsLab/neuroscout)
 
-To set up docker, ensure docker and docker-compose are installed.
+This is the repository for the neuroscout server.
+
+Requirements: Docker and docker-compose.
 
 ## Seting up environment / config variables
 First, set up the main environment variables in `.env` (by copying .env.example).
--- a/README.md
+++ b/README.md
@@ -8,11 +8,11 @@ This is the repository for the neuroscout server.
 Requirements: Docker and docker-compose.
 
 ## Seting up environment / config variables
-First, set up the main environment variables in `.env` (by copying .env.example).
+First, set up the main environment variables in `.env` (see: `.env.example`).
 Set `DATASET_DIR`, `KEY_DIR`, and `FILE_DATA` to folders on the host machine.
 
-To set up environment variables for pliers (optional), modify `.pliersenv.example`,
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -3,70 +3,62 @@
...
-In addition, configure where file keys (such as for Google) are located:
+Optionally, set up pliers API keys for feature extraction in `.pliersenv` (see: `.pliersenv.example`).
+[More information on pliers API keys](http://tyarkoni.github.io/pliers/installation.html#api-keys)
 
-    KEY_DIR=/home/me/.keys
-    export KEY_DIR
+Next, set up the Flask server's environment variables by modifying `neuroscout/config/example_app.py`
+and saving as `neuroscout/config/app.py`.
 
+Finally, set up the frontend's env variables by modifying `neuroscout/frontend/src/config.ts.example`
```
Shotgunosine: Might want to mention here that they'll need to set up a google sign-in project here:
https://developers.google.com/identity/sign-in/web/sign-in

Shotgunosine:
Yeah looks good other than the comment I left. I might try changing the settings to point the the remote host i've got it on and see if it just works for me. If so, those instructions might also be nice to add to the readme somewhere.

Shotgunosine:
Alright, so if you want to set it up on a remote server you need to:
1) get a google api client id that corresponds to that server
2) update `neuroscout/frontend/src/config.ts` to reflect the remoter's server_url
3) update `SERVER_NAME` in `neuroscout/config/ap.py`
4) update `server_name` in `nginx/nginx.dev.conf`

neuroscout/requirements.txt
```diff
@@ -46,8 +46,8 @@ Pillow==4.1.1
 psycopg2==2.7.1
 progressbar2
 py==1.4.33
-grabbit==0.2.3
--e git+https://github.com/bids-standard/pybids.git#egg=pybids
+grabbit==0.2.6
+pybids=0.7.0
```
Shotgunosine: That should be an ==

Shotgunosine: Also, that still produces this warning:
`pybids 0.7.0 has requirement grabbit==0.2.5, but you'll have grabbit 0.2.6 which is incompatible`

tyarkoni: 0.7.0 *doesn't* require 0.2.6, it requires 0.2.5. The 0.2.6 update has been merged into master, but will be in 0.7.1. So probably the solution is to remove the explicit grabbit dependency and let pybid do its thing.

neuroscout/requirements.txt
```diff
@@ -46,8 +46,8 @@ Pillow==4.1.1
 psycopg2==2.7.1
 progressbar2
 py==1.4.33
-grabbit==0.2.3
--e git+https://github.com/bids-standard/pybids.git#egg=pybids
+grabbit==0.2.6
```
tyarkoni:
```suggestion
```

PR #462, opened 2019-01-24
Add auto-fill functionality

adelavega:
WIP.

Adds a route at `/api/analyses/{ID}/fill`.
This is a `POST` route. It will try to fill sensible defaults for the following fields:
- `runs`: Defaults to all runs
- `predictors`: Will look in model and try to find names from `X`. Then will query for all the latest predictors for the specified runs matching those names.

Note that this is not a `PUT` route, so changes to the model (e.g. blanking `predictors`) will have to be done using the appropriate route.

Closes #462

### DIFF
--- a/neuroscout/core.py
+++ b/neuroscout/core.py
@@ -2,19 +2,7 @@
 """ Core Neuroscout App """
 import os
 from flask import Flask, send_file, render_template, url_for
-from flask_mail import Mail
-from flask_caching import Cache
-from flask_jwt import JWT
-from flask_cors import CORS
-from flask_security import Security
-from flask_security.confirmable import confirm_email_token_status, confirm_user
-from auth import authenticate, load_user, add_auth_to_swagger
-from models import *
 from database import db
-
-from apispec import APISpec
-from flask_apispec.extension import FlaskApiSpec
-from utils import route_factory
 
 app = Flask(__name__, static_folder='/static')
 app.config.from_object(os.environ['APP_SETTINGS'])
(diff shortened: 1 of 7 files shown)


adelavega:
Allright @tyarkoni this seems to be working. Are you OK with the route being only a "fill" and not a `PUT` / edit route? It's cleaner to me even if takes two API calls. 

Also, for partial fill, I was thinking the model should be edited to reflect only the names it could find. Ideally, it would also remove transformations/contrasts, etc that aren't valid as well... 

Todo:

- [ ] Add tests / really ensure "newest' query is working (had problems in the past with it)
- [ ] Edit model in partial fills

tyarkoni:
NIce! Yeah, that sounds fine to me.

Re: transformations and contrasts, I think if any are invalid, we should remove all of them, so that it's obvious to the user. The alternative risks making subtle changes that go unnoticed and/or change the meaning of the resulting model in unpredictable ways.

adelavega:
Sounds good. I think searching for input names on transformations or contrasts should be easy enough. This would be nice because then if the only changes are confounds (e.g. # of noise components, not convolved with HRF which is a transformation) all else would stay the same.

adelavega:
@rwblair Can you implement the following changes to the frontend:
- If there is a change in dataset, and model is not empty (i.e. only after "cloning" or if someone get through making a model and goes back to the first page), then do the following:
- Let users select the task/runs (we won't use the autofill for the runs on the frontend, as "task" needs to be selected, and autofill would select both tasks
- Upon trying to navigate past the first tab, remove all `predictors`, `POST` to `/fill` and update the anaysis object. 

Let me know if this makes sense. Probably would require keeping track of which dataset was selected when first navigating to the first tab, and if it changes when navigating out, then do a `/fill`

adelavega:
I think this is ready to merge. @rwblair you can add the frontend changes in a separate PR, and if any bugs come up we can fix them there.

Issue #442, opened 2018-12-20
Auto-save to server when you change tabs (i.e. next/back)


adelavega:
Like when going to the Review tab

adelavega:
But without the pop up that analysis has been saved.

Issue #429, opened 2018-12-17
Run filter issues


rwblair:
clear filters to reselect all instead of select none.

Toggling select all twice causes the number of selected to be shown as "all selected", should show the correct count.

PR #476, opened 2019-01-30
fixes for #442 and #429

rwblair:
also includes changes from recently closed PR to fix first design matrix opening by default.

### DIFF
--- a/celery_worker/tasks.py
+++ b/celery_worker/tasks.py
@@ -165,7 +165,9 @@ def _plot_save(dm, plotter, outfile, **kwargs):
 class PathBuilder():
     def __init__(self, outdir, domain, hash, entities):
         self.outdir = outdir
-        self.domain = domain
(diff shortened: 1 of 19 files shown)


rwblair:
Tested the command that failed in travis: `python -c "import imageio; imageio.plugins.ffmpeg.download()"`  locally and it worked, so the ffmpeg url is still up and active.

adelavega:
So travis is just acting funky then?

rwblair:
I think so, its strange because the ffmpeg file is hosted on github.com, and we are able to clone from github.com seems like it shouldn't be a connectivity issue.

Guess I'll wait a little bit and try re-running see if it clears up. Wish travis had ssh access.

rwblair:
@adelavega ok I think I have fixed that, and another issue of applying filters one after another that I must of introduced.

adelavega:
Sweet

PR #481, opened 2019-02-05
Add 404 page and have /api/ go to swagger ui

rwblair:


adelavega:
Odd-- doesn't seem to work still

adelavega:
https://alpha.neuroscout.org/api

PR #482, opened 2019-02-05
Speed up API

### LINKED ISSUE #475: Profile/speed up API
Generally, we need to profile various API routes (even though not heavily used by the frontend, which has been the primary focus thus far), and find major performance problems.

Known performance problems right now are:

- Slow serialization with Marshmallow. We addressed this in #293 for `/compile` with some success (although, it may still time out-- need to check on that). Should address this for `/api/predictor-events` and maybe a few other routes (at least to dump, if not ingest)
- Computing stats on `/predictors`. This could be slow, worth looking into.
- Getting newest predictors (`group_by` `func.max`). Maybe can do this in raw SQL.

adelavega:
WIP. Closes #475 

Removing dynamic non-lazy joins and on the fly stats on Predictors took the response of `/api/predictors` from 38s to 3s.

### DIFF
--- a/neuroscout/models/analysis.py
+++ b/neuroscout/models/analysis.py
@@ -51,7 +51,7 @@ class Analysis(db.Model):
     # If cloned, this is the parent analysis:
     parent_id = db.Column(db.Text, db.ForeignKey('analysis.hash_id'))
 
-    results = db.relationship('Result', backref='analysis')
+    results = db.relationship('Result', backref='analysis', lazy='dynamic')
     predictors = db.relationship('Predictor', secondary=analysis_predictor,
                                  backref='analysis')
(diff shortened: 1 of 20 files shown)


adelavega:
I will consider reviving `dynamic='lazy'` joins if necessary. I don't think they were the major source of the problem, as this was a lazy load. The main difference is that lazy dynamic returns a query object that can be further filtered, whereas `lazy='select'` returns a SELECT `.all()` python list of objects.

rwblair:
@adelavega I did not have any issues loading predictors on the front end with my local instance. I'm  doing a fresh restore on my database from that dump you gave me and its taking a minute, so I can test more thoroughly again tomorrow if need be.

PR #485, opened 2019-02-06
Ignore query string order. Closes # 473

### LINKED ISSUE #473: Memoization should ignore invalid parameters
Currently, if you request https://alpha.neuroscout.org/api/predictors?run_id=333 twice, the second query is cached. But if you then request https://alpha.neuroscout.org/api/predictors?run_id=333&boo=1, the cache is ignored. The backend should only use a restricted set of valid parameters in memoization, otherwise we're going to eat up clock time and disk space with redundant results.

adelavega:
Closes #473 

I'm removing our custom cache key function, and replacing it with the build in query string support. This will ignore the order of the arguments. 

Actually ignoring invalid arguments is not as easy, because the libraries that parse the query string, and cache don't talk. We could add this functionality but I honestly think its not necessary because its rare we'll get so many invalid requests (especially if using Swagger or `pyns`) as to cause a problem. In addition, flask-caching handles the disk size of the cache, so unless that becomes a problem, I think we are OK.

### DIFF
--- a/neuroscout/resources/dataset.py
+++ b/neuroscout/resources/dataset.py
@@ -31,7 +31,7 @@ class DatasetSchema(Schema):
 
 class DatasetResource(MethodResource):
     @doc(tags=['dataset'], summary='Get dataset by id.')
-    @cache.cached(60 * 60 * 24 * 300 query_string=True)
+    @cache.cached(60 * 60 * 24 * 300, query_string=True)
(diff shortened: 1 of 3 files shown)


adelavega:
If we decide this is still a problem, I could write a custom wrapper to both `webargs` and `flask-caching` that would filter the valid key arguments based on the web arg arguments.

Issue #468, opened 2019-01-29
Only passed analyses should be cloneable


adelavega:
The API respects this, but the frontend shows a Clone button for FAILED as well as DRAFT.

Issue #421, opened 2018-12-05
Add consistent headings for each tab


adelavega:
Not all tabs have a heading (like "transformations" tab does). Ensure all tabs have a header at the top to orient users.

Issue #408, opened 2018-11-25
Failed analyses should be editable


adelavega:
Currently it asks you to clone and edit, but if it fails, it should go back to edit mode

rwblair:
When loading a failed analyses land on run tab with run errors, and ability resubmit in addition to being able to go back to other tabs to edit.

adelavega:
Closed by #493

PR #493, opened 2019-02-12
Random bug grab bag.

rwblair:
Includes fixes for #421 #468 and #408


adelavega:
Cool, looks good. I added a small backend change to go along.

PR #494, opened 2019-02-12
update antd to fix transparent dropdown menus.

rwblair:
#fixes 465

There is an issue with some @types packages keeping their own version of react types. I had to forcefully remove @types from node_modules to get this to work.


adelavega:
Great!

PR #504, opened 2019-02-20
443 389 preload predictors consistent form err

### LINKED ISSUE #389: Make frontend form validation more explicit to user.
Traditionally html form validation will complain when a form is being submitted. Need to trigger this action when the next tab button is hit.

### LINKED ISSUE #443: Immediately load available predictors when reloading an existing in draft analysis.
Otherwise jumping to transform tab will lead to wonky behavior where we have selected predictors but not any available predictors.

rwblair:
fixes #443, set another flag in the main store for the initial load of predictors. We want predictors to be loaded the first time we see runIDs populated, which is several re-renders deep from the constructor since we wait on the api.

fixes #389 consolidated the code that displays form validation error messages to be consistent across tabs that display them, just transforms and contrasts right now.

### DIFF
--- a/neuroscout/frontend/src/Builder.tsx
+++ b/neuroscout/frontend/src/Builder.tsx
@@ -240,10 +240,6 @@ export default class AnalysisBuilder extends React.Component<BuilderProps & Rout
           return data;
         })
         .then((data: ApiAnalysis) => {
-          // tslint:disable-next-line:no-console
-          console.log('maybe here?');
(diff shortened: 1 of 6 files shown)


adelavega:
Looks good. One more minor change to bundle. Title for add new contrasts is now : 'Add a new contrast:'
Remove the single quotes please.

Otherwise in my limited testing LGTM

PR #505, opened 2019-02-21
shorten author list in dataset table, offer tooltip with full list

### LINKED ISSUE #501: Shorten author list
For aesthetic purposes, it would be nice to shorten the author list such that if there are more than 3 authors, only the first and last are shown, with "..." in between.

Minor easy issue.

rwblair:
fixes #501 if we assume authors will never have comma in their name. There's another solution if we think they will, but we might have to track the short auth description and long description at the same time.


rwblair:
Eh just took a little bit of work to make it work even if a name might have a comma, build the authors list as a single string when its being displayed instead of right when we get it from the api, so then we don't have to run a split by ',' on it in the display like my first commit.

adelavega:
Can you make it so instead of this: `E.S. Finn,...,P.R. Corlett`
its this: `E.S. Finn ... P.R. Corlett`

Spaces instead of commas around the ellipses

PR #518, opened 2019-03-14
Have frontend make use of /fill api end point.

### LINKED ISSUE #508: Add autofill / smarter cloning to frontend
Currently there is a "fill" route, but it looks like frontend doesn't use it. 
Relatedly, when switching datasets, the subjects are not cleared out (after cloning)

rwblair:
fixes #508.

I have turned on pagination for predictor selection for performance in testing. Its using defaults, I can set it to have 100 or 200 elements per page if we want, or if there is another way we can think of  organizing and making predictors explorable.

The /fill endpoint is only called after we change datasets and switch tabs.


adelavega:
LGTM. The only minor commment I have is maybe a little spinny loading thing would be nice when the frontend is hitting the `/fill` route

PR #520, opened 2019-03-19
Add endpoint to receive neurovault results

### LINKED ISSUE #196: NeuroVault integration
Neuroscout CLI should be able to push fitted images to NeuroVault seamlessly.

First, how to handle authentication? 
 - Ideally, if your account is the same email as NeuroVault (especially if you used single sign in with say Google) you shouldn't need to authenticate to push to NeuroVault. However, this probably requires an integration step to pre-authenticate users w/ NeuroVault. We could automatically do this for all matching email addresses automatically though. 
 - A less optimal solution is to ask for user name and password at the command line and cache authentication token once we get it from the NeuroVault API

Also, does NeuroVault API have push routes? How easy is it to push?
Ideally there should be a lightweight python library for NeuroVault API

For reference, see how Neurolearn integrates with NeuroVault, and talk to @chrisfilo

adelavega:
Closes #196


adelavega:
Upload to NeuroVault is functional.

Left to do:

- [ ] Decide if current upload model makes sense (+1 collection per analysis, with hash_id + datetime name)
- [ ] Tests
- [ ] Update pyNS to upload

adelavega:
@tyarkoni minor question about naming / number of uploads. 

Do you foresee that we might upload the same analysis twice? Given that fitlins versions can change, and runtime options may change the execution, I set this up so that each analysis can have multiple NeuroVault uploads/collections associated with it. 

Since NeuroVault can only accept one collection of each name, I set the collection name to be hash_id + timestamp to be unique. E.g.: https://neurovault.org/collections/4996/

Seem OK?

tyarkoni:
Under this scheme, what would happen if a user runs the exact same bundle (same version of all dependencies) twice? What we want to avoid is a situation where someone (or possibly different people) just re-runs the identical analysis and we have multiple versions sitting in NeuroVault without being able to distinguish between them. If it's feasible, I'd prefer to compute a hash based on the unique analysis AND anything we think we might care about, and not based on a timestamp. Then we can check NeuroVault to see if a collection with the specified name already exists before uploading.

Other than command-line options and fitlins version (and probably nistats as well), is there anything else we'd need to worry about? If that's all, maybe we could just add a hashing utility to fitlins that returns something we can combine with the analysis ID? @effigies, any thoughts?

effigies:
That seems reasonable. I think `(fitlins.__version__, nistats.__version__)` would be a reasonable start. As dependencies grow, just add to the list of things hashed. One question is how long a hash you want.

adelavega:
Maybe fitlins can embed a hash of its CLI + versions in `dataset_description.json`?`.

That said, ideally the command line options to neuroscout at least don't allow the user to change the analysis much, if it all. So the only thing that *could* change is versions. 

In 95%+ of cases, I think a single run/upload per hash is what we'd like, and will happen. So actually, I think I lean towards --force-neurovault for simplicity.

effigies:
Maybe it would make more sense to just package version info in `dataset_description.json`, which can be hashed by neuroscout.

adelavega:
I think that should definitely happen either way, yeah. Although I think I'll go with the date approach for now.

adelavega:
This is working and pyNS is updated, so I'm going to merge this for now and we can change later if we find a better approach. 

@rwblair I think we should add a little indicator pointing to NeuroVault images if they are available. We can discuss more later about how specifically.

PR #523, opened 2019-03-21
Fix slow predictor table rendering.

rwblair:
First commit enables pagination and sets the page size to 200.


adelavega:
OK looks good. If there are multiple pages, does search work across them? if so, LGTM you can merge!

PR #528, opened 2019-03-28
display dataset name in analysis list, display additional dataset information on review page.

### LINKED ISSUE #498: Display dataset info on Review page
This is related to #449, but probably merits its own issue. Currently when reviewing an existing analysis, it doesn't seem like there's an easy way to tell what dataset was used. I know that this isn't part of the BIDS-StatsModel spec, so maybe it shouldn't be shown in the accordion, but it does need to be displayed somewhere.

rwblair:
fixes #498


adelavega:
Looks good, but for some reason the `Dataset` does not display in "myanalyses" only in public analyses. 

When I try to sort by Dataset in myanalyses, I get the following error:

```
TypeError: "e.datasetName is undefined"
```

PR #530, opened 2019-03-28
Generate vega-lite plots for reports

### LINKED ISSUE #525: Use altair / vega-lite for report plots
Altair looks awesome and compiles to vega-lite, which compiles to vega, which compiles to d3. 
Seems like a nice way to produce interactive plots from python -> web

adelavega:
Closes #525 

Need to update frontend and convert all plots. Also need to further refine the plots.


adelavega:
Allright, I added the correlation matrix, and it looks decent (although I couldn't do the calculations in altair, and they had to be done prior to vega/altair).

Here's some outstanding issues:
Major:

- [ ] `yarn build` results in buggy build. @rwblair is on it
- [ ] I think it would be better to organize the Collapse by subject, rather than by plot type. Especially given that we should add a dialog to select which run to generate a report on. Thus, the vega component shoudl look over both plots at the same time, and I should include which subject/run combo each result is

Minor:
- [ ] Vega/altair is still a little buggy. For example, I can't for the love of god get the x-axis limits correctly
- [ ] I'm not super happy w/ the interactivity so I'm going to keep messing with it a bit. It would be nice if you could unselect, and I should probably add a legend to the line plot. 
- [ ] Maybe I can add some tool tip hints / instructions

adelavega:
Hey @rwblair I reorganized the Vega components a bit, so they are tabbed by subject. What would be awesome is if we could also take care of #533 and add a button that lets users change which runs are request in a report. This would then fill the appropriate tabs titles. 

My question is if you want me to return a matching list with the entities for each response. Or if I return them in the order you asked (e.g. run_id = 333, 334, 335), you can then get the entities for those runs, and fill int he Tab with the appropriate formatting (e.g. "Sub 123, Run 1").

adelavega:
In terms of status: I'm pretty happy with the plots I'm putting out. Although, beware a couple things:
- For 1 subject, it returns about 1MB of JSON data... That's the downside of these type of plots
- I'm seeing a weird bug that it seems to remove the HRF transformation sometimes. I hadn't noticed before, but it became pretty obvious because I'm using that transformation to sort the columns in the plots (e.g. columns that got convolve go first, as they are presumably "of interest"). Can you see if you can replicate that? It was quite unnerving. 

I think we take care of those issues + obviously the production build issues, this PR will be ready to go, and Reports will be looking very awesome.

adelavega:
This also closes #531

adelavega:
@rwblair can you do a final review and see if you can coax travis into passing? otherwise it LGTM

PR #542, opened 2019-04-17
Only create Predictors for EFs present in 90% of runs

### LINKED ISSUE #541: Disable Predictors not present across all runs by default
Given that:
1) Fitlins can't handle empty predictors in a run
2) For some extractors there are potentially way too many features (1000s)
3) The frontend struggles a bit (at least in terms of UX) with so many features

I suggest that by default, when extracting, if a feature is not present in all runs, it is not `active` by default. We could de-active at the feature or predictor level (e.g. not create a predictor at all, or create it, but not display it by default. Given that it takes a long time to create features, I would at least suggest for those not present in 50-80% of runs, they are not even created.

adelavega:
Closes #541 

I decided on a 90% default

That code in general is not the most efficient and a bit hard to debug, but I can't think of a much better way to do it for now.

At least this will speed things up by only processing useful Features (i.e. those in most runs).

### DIFF
--- a/neuroscout/config/transformers.json
+++ b/neuroscout/config/transformers.json
@@ -28,7 +28,7 @@
     ["MeanAmplitudeExtractor", {}],
     ["SpectralCentroidExtractor", {}],
     ["VADERSentimentExtractor", {}],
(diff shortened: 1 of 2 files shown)


adelavega:
This seems to be working well

PR #547, opened 2019-04-26
Compile bilbiography for analyses

### LINKED ISSUE #543: Citation reports
Add a route `/api/reports/bibliography` which compiles a list of citations for the Dataset + Extractors/Features used + (c) the tools used in any part of the workflow (pliers, fitlins, etc.)

This could be a JSON of DOIs or another standard format for citations if one exists that is not a huge pain to create.

To keep track of citations, we need to add this information for each extractor to `feature_schema.json` (to associate features w/ citations)

A table for citations/dois could be created, with has a many-to-many association with EFs.

adelavega:
Closes #543 

Looks like `CSL` is a good JSON standard for citations. In a .json file, I will keep a list of citations, with keys representing either `tools`, `datasets` or `extractor_names`. I don't think there should ever be clashes.
Then, an API endpoint will simply take a base set of tools + the dataset_name + extractor_names in an analysis, and return a `CSL-JSON`.

We could additionally try using citeproc-py to render a bibliography in APA or other format. However, the python library looks a bit clunky and not actively maintained, so I'm weary of using it. Also, this limits the formats we can provide (as it would require going back and forth, and doing it only on compilation).

Instead, I think its best if we use `citations.js`, a much more actively developed library, to render / compile the bibliography. This way, users could select a format, and compile it, including BibTex or another common format.

### DIFF
--- a/neuroscout/Dockerfile
+++ b/neuroscout/Dockerfile
@@ -4,11 +4,6 @@ ARG DEBIAN_FRONTEND=noninteractive
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
-COPY requirements.txt /usr/src/app/
-RUN pip install --no-cache-dir -r requirements.txt
-
-COPY . /usr/src/app
-
 RUN apt-get -qq update
 RUN apt-get install -yq ffmpeg tesseract-ocr
 RUN pip install -e git+https://github.com/tyarkoni/pliers.git#egg=pliers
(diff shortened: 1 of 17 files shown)


adelavega:
To add more detail as to how `bibliography.json` is being populated. Zenodo natively exports to `CSL-JSON`, which is nice. For papers, I am importing to a local Zotero library, which can then be easily exported to `CSL-JSON`.

adelavega:
image

This is almost done. @tyarkoni @rwblair what do you think?

Unfortunately, I could not get `citation.js` to work, so instead the API returns APA formatted strings. That package seems to be in active development, so I'm hopeful in the near future I can instead pass CSL-JSON, which the frontend can then convert on the fly to BibText, and various Text formats.

adelavega:
I modified `bibliography.json` so that the second key refers to the feature name for extractors. `.*` matches all. This is for extractors like the `PredefinedDictionary` which have features with different references each. 

For datasets, and tools, they always have a `.*` defined, and that is what is always matched.

tyarkoni:
Looks great!

PR #551, opened 2019-05-01
Bump pybids 0.9 - sqlalchemy

adelavega:


adelavega:
Need to resolve: https://github.com/bids-standard/pybids/issues/451

PR #553, opened 2019-05-08
Enh/add new variables

adelavega:
Add new `subtlexusfrequency` variables


tyarkoni:
Looks like the description is the same for all variables... I assume you're planning to pull that directly from the `dictionaries.json` file in pliers, but just flagging it in case it's an oversight.

adelavega:
Oh goddamn it

adelavega:
OK That was pretty easy. Fixed.

PR #558, opened 2019-05-11
Ingest Rev API Text

adelavega:
- Minor changes to successfully import files from Rev API. 
- Extract more text related features.

### DIFF
--- a/neuroscout/populate/convert.py
+++ b/neuroscout/populate/convert.py
@@ -164,6 +164,9 @@ def ingest_text_stimuli(filename, dataset_name, task_name, parent_ids,
                          variable play speeds.
     """
     df = pd.read_csv(filename, delimiter='\t')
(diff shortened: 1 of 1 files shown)


adelavega:
After looking at the fourth run transcripts, I think I will try to force align it with manually given time stamps, as the transcript is mostly good. 

Also worth checking the other 3 runs carefully.

adelavega:
Okay, that was a pain but it is done,.

adelavega:
To add some documentation. To get the transcripts from rev-api, use my fork: https://github.com/adelavega/rev-api

and get orders using: 

```
client.get_orders_page()
client.save_order_transcripts(order_number, format='tsv', base_path='/path/to/folder')
```

These transcripts can then be ingested into neuroscout using:
`ingest_text_stimuli`

PR #574, opened 2019-05-20
Add db to celery

adelavega:
This is a refactor of the models in neuroscout so that they use vanilla `SQLAlchemy`. This was they are importable from other non-Flask places, such as Celery.

Currently, relationships are problematic, they me be a difference is now ForeignKeys are defined between flask-sqlalchemy and vanilla.

```
sqlalchemy.exc.NoForeignKeysError: Could not determine join condition between parent/child tables on relationship ExtractedFeature.generated_predictors - there are no foreign keys linking these tables.  Ensure that referencing columns are associated with a ForeignKey or ForeignKeyConstraint, or specify a 'primaryjoin' expression.
```

### DIFF
--- a/neuroscout/auth.py
+++ b/neuroscout/auth.py
@@ -1,13 +1,17 @@
 """ Auth related functions """
 from datetime import datetime
-from models.auth import user_datastore, User
+from database import db
+from models import User, Role
 from flask_security.utils import verify_password
+from flask_security import SQLAlchemyUserDatastore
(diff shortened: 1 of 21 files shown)


adelavega:
This approach is not going to work because so many of my libraries depend on `flask-sqlalchemy` and using vanilla `sqlalchemy` models will lose lots of functionality, and require a lot of re-writing of code. 

Instead, will opt for #575 which is to use the flask context from the main app to instantiate the db/models.

PR #575, opened 2019-05-20
Integrate celery to flask context for db

adelavega:
Alternative to #574 

This approach requires making `neuroscout` an installable package, so that I can pip install it from the celery  instance and import the models/db/app_context from flask. The main challenge has been that the relative imports in the package are all wonky, and its hard to fix them all (although a good idea nonetheless!)

Gunicorn now instantiates from one level higher, so that `neuroscout` is seen as a package.


adelavega:
This should be working now, need to test more, and on live server. Unfortunately, tests for celery may no longer worker, because flask uses a temporary way of saving info the db that is not easily shared with a different instance. Local tests mail fail in those instances, may want to remove them for now.

adelavega:
This is looking good finally.

@rwblair btw, note that in `.env` `APP_SETTINGS=neuroscout.config.app.DevelopmentConfig`

adelavega:
Still need to test upload, will do when testing CLI tomorrow

adelavega:
Tested compiling an analysis and uploading results and it works now. 

Merging

PR #577, opened 2019-05-23
576 fix no task stat model

rwblair:


rwblair:
fixes 576, not 526 like branch and commit say

adelavega:
LGTM

PR #584, opened 2019-06-02
[WIP] Link to neurovault uploads.

rwblair:
Currently will only display latest failed attempt with traceback, and will only show latest pending attempt with traceback if there are no successes. All success shown if available. Not sure how to word the failed and pending titles.

API endpoint currently only hit on initial analysis load and when the analysisId changes.

### DIFF
--- a/neuroscout/frontend/src/Status.tsx
+++ b/neuroscout/frontend/src/Status.tsx
@@ -294,12 +294,14 @@ export class StatusTab extends React.Component<submitProps, statusTabState> {
           })
         }
         {(this.state.nvUploads.failed) &&
+          <p>
           <Alert
             message="Last failed upload:"
             description={`Failed at ${this.state.nvUploads.failed.uploaded_at}
               <br/>${this.state.nvUploads.failed.traceback}`}
             type="error"
           />
+          </p>
         }
         </div>
       }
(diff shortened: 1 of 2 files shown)


adelavega:
Nice! Thanks! This looks great. I may push a change with minor aesthetic tweaks but otherwise I think this is what we need. 

Question: what do you mean by:

> API endpoint currently only hit on initial analysis load and when the analysisId changes.

What if you're looking at `Status` and then an upload gets pushed. Maybe we can have a little "refresh" icon? Or I guess it's probably reasonable for people to expect to have to refresh the page to get the latest?

adelavega:
Minor (more of a note to self): Let's format this timestamp nicer `Uploaded at: 2019-05-27T19:03:00`

adelavega:
If there is nothing to show, either have  a place holder below "NeuroVault Uploads", or don't show. The latter is probably best, especially if it hasn't been compiled yet. Maybe after compilation it can have a place holder

rwblair:
Unfortunately Date.parse in the app wasn't able to parse '2019-05-27T19:03:00' (though my local node install was). Need to find another way of parsing (or can start just replace that inner T with a space? or remove the time part all together?)

adelavega:
Replacing T with a space should be fine, I think!

adelavega:
Is this still WIP? Looks pretty good to me.

adelavega:
I'm going to merge, it looks good to me

PR #587, opened 2019-06-04
filter out contrasts whose source is 'fmriprep' in autocontrast creation

### LINKED ISSUE #506: Only create auto contrasts for non confounds
Confounds should only be for predictors that are not "fmriprep" in type

rwblair:
fixes #506


adelavega:
LGTM. Can you also do the same for HRF selection?

Issue #593, opened 2019-06-08
Disable some error prone transformations for now


adelavega:
Until we can better report back what went wrong to the user. 

E.g. Replace, etc...

PR #597, opened 2019-06-08
Fix normalize dm, and remove some transforms

### LINKED ISSUE #581: Not all variables are scaled
It seems that some fmriprep variables are not scaled in the reports desing matrix plot:

image

adelavega:
Closes #581 and #593


adelavega:
Sweet, looks good.

PR #607, opened 2019-06-25
Neuroscout tour prototype

adelavega:
I was thinking about how to document Neuroscout, and I realized that I personally never read frontend manuals unless I absolutely have to. I still think it's a good idea, especially for more technical model building details, but I think if we intersperse tours/tips/hovers throughout the interface to answer most questions, more people will actually see it.

I messed around with `reacttour` and it was pretty easy to make a tour that could be shown to first time users. The code is quite compact, and easy to trigger actions including Router changes. It gets a bit trickier once the builder starts (arguable the most important part), but doable. 

Just an idea, but here's some screenshots of what it looks like. 

Screenshot from 2019-06-25 15-58-29

Screenshot from 2019-06-25 15-58-40

Screenshot from 2019-06-25 15-58-47

Screenshot from 2019-06-25 15-58-55

Screenshot from 2019-06-25 15-59-04


tyarkoni:
That looks good to me. I think a manual is still important, but it's fine if it focuses more on the implementation details (i.e., stuff people need to know for their methods sections) than on the UI.

adelavega:
Yep, agreed!

adelavega:
Okay, for this I added tooltips to all the pages.

We still need to:
- [ ] Implement tooltips only pop up on first time
- [ ] Check text for all tooltips, maybe add some for every transformation
- [ ] Decide if tour is necessary at all, or if we should take it further. I think it might be OK as is, if the pop ups follow the tour.

adelavega:
I'm going to go ahead and merge this, and we can keep working on it.

In particular, we need to add a place in the `Transformations` definition where we can put the tooltip text, and add that tooltip (with the idea each transformation would have its own tooltip helper).

We also need to add back the Predictions form (see #613)

PR #608, opened 2019-06-25
Avoid re-doing report for minor post-DRAFT changes

### LINKED ISSUE #585: Avoid redoing report for PASSED analyses
One issue I seem to be having is that looking at an old analysis triggers a new report generation. I wonder if this is because after compilatiion the "changed" timestamp changes, and the old report is "stale". 

Maybe we need to either avoid this, or ensure a final report is triggered with compilation. Mostly would be nice for people browsing public reports not to have to wait to see the design matrix. It makes the app seem broken.

adelavega:
Fixes #585 

What was actually happening is that I changed the name of the analyses post-compilation. 

The easy fix here is to not change the `modified_at` timestamp for trivial changes post-compilation, such as `name` `description` or `private` status.


adelavega:
OK this works. Easy.

PR #611, opened 2019-07-03
Add predictor/collection upload route

### LINKED ISSUE #603: Interface for uploading for custom annotations
Related to #589 , we need a general way to upload custom annotations for datasets. 

For `ParanoiaStory` there were also some annotations that @esfinn said she had. If users were able to upload these through the web interface, that would be ideal.

adelavega:
- Closes #603 by adding a route to upload custom predictors
- Fixes an issue where the API Docs did not indicate file type correctly
- To fix the above, and to better organize the core app, I separate the API docs into its own file, which can them be imported from other modules without circular issues. The resulting core.py is also much cleaner, and I reorganized the imports so that only three need to be sequentially later in the file (because of complicated imports, again).

Note: because of sequential db changes, this depends on #607 to be merged.


adelavega:
The upload route is working (still need to test Celery task).

The correct request looks like:

```
curl -X POST "http://localhost/api/predictors/create" -H  "accept: application/json" -H  "authorization: JWT [JWT]" -H  "Content-Type: multipart/form-data" -F "dataset_id=9" -F "runs=258,256" -F "runs=252,251" -F "event_files=@/path/run_1.tsv;type=text/tab-separated-values" -F "event_files=@/path/run_2.tsv;type=text/tab-separated-values" -F "collection_name=test"
```

adelavega:
Note the repetition of parameters for runs and event files, but comma separated values for runs

adelavega:
Upload route is working!

Here is a summary.

1) Make a `POST` request to `/predictors/create` with your JWT.
The list of runs (and run IDs) must equal the number of files being uploaded. 
Any given run can only be associated with a single event file. 
Event files can have multiple new `Predictors`, but all of the event files uploaded at once but have the same columns.

Example:
```
curl -X POST "http://localhost/api/predictors/create" -H  "accept: application/json" 
-H  "authorization: JWT [JWT]" -H  "Content-Type: multipart/form-data" 
-F "dataset_id=9" -F "runs=258,256" -F "runs=252,251" 
-F "event_files=@/path/run_1.tsv;type=text/tab-separated-values" -F "event_files=@/path/run_2.tsv;type=text/tab-separated-values" 
-F "collection_name=test"
```
Returns:
```
{
  "collection_name": "test", 
  "id": 13, 
  "predictors": [], 
  "status": "PENDING", 
  "traceback": null, 
  "uploaded_at": "2019-07-05T19:3"
}
```

This creates a new `PredictorCollection` which (like `NeurovaultCollection` or `Report`), is used to track the Celery status of the upload, and also serves to group together the new predictors that were uploaded. This is also associated with  your us id.

To check the status, make a GET request:

`curl -X GET "http://localhost/api/predictors/create?id=11" -H  "accept: application/json"`

which might return something like this if it succeeds:
```
{
  "collection_name": "test", 
  "id": 11, 
  "predictors": [
    {
      "id": 19244, 
      "name": "reaction"
    }
  ], 
  "status": "OK", 
  "traceback": null, 
  "uploaded_at": "2019-07-05T19:3"
}
```

You can now use those predictors like normal, and they will display for all users (maybe this is not what we want to do? might want to have this as an option private/public?)

To see which collections you've created do a `GET` on `/api/user` with your JWT:

```
{
  "email": "user@example.com", 
  "first_login": null, 
  "name": null, 
  "picture": null, 
  "predictor_collections": [
    {
      "collection_name": "test", 
      "id": 11
    }
  ]
}
```

adelavega:
WDYT? @rwblair 

Does this seem reasonable for the frontend?

Minor design quibbles:
- Maybe this should be under a different route? Maybe it makes more sense to have `/predictor_collection/` and do a `POST` and `GET` to that? rather than `/predictor/create`. It would be more "RESTful". OTOH, this is more consistent with how we handle neurovault uploads and reports, so I think it's probably good as is. 
- It might also make more sense to think of "PredictorCollection" is the appropriate name. 

To do still:
- [ ] Should we abstract logic for creating Predictors a bit, and import from worker?
- [ ] Add tests
- [ ] Check that clearing cache works

adelavega:
I added some basic tests. Can't fully test because Flask doesn't write out to the test_db, so Celery can't see it.

At a separate time, we should fix this to allow for real testing of Celery tasks.

Issue #612, opened 2019-07-08
Move modal and menu bar components out of app.tsx


rwblair:
Help de clutter the file mainly responsible for routing.

Issue #610, opened 2019-06-28
Add Google analytics


adelavega:
Currently, tracking usage is not so easy.

Maybe we should add Google Analytics (easy) and/or add an internal usage tracker dash (that could give more detail info, such as number of new analyses, etc...)

Issue #569, opened 2019-05-15
Analyses status not updating


adelavega:
Seems like the status stopped updating.

I can add a comma separated list to GET `compile` if that will help check statuses easier.

adelavega:
To clarify what is happening: The polling seems to be broken. A hard page refresh will get it to display.

We need to fix the polling, and/or add a refresh button + force refresh on navigation.

PR #616, opened 2019-07-10
Add analytics, reorg front end code

rwblair:
Lot of things got moved around. Need to test some more to be certain everything actually works as expected.


rwblair:
@adelavega Heads up I was running into performance issues with the react debugger running in firefox with large numbers of predictors. Let me know if you notice any other wonky performance issues.

Also need to add a 'ga_key': 'UA-whatvever' to the config.ts for analytics to work. I've been seeing my computer come up in the dashboard when using the sites.

rwblair:
Spent some time this morning profiling the site, the slowness with selecting/unselecting elements on a long list of predictors can't really be avoided. It does re render each line in the table, which only takes a few ms per row, but at a max of 200 rows it adds up.

Not for this PR, but what do we think about putting each category of predictor into its own table that starts out collapsed?

rwblair:
fixes #612 #610 and #569

Issue #620, opened 2019-07-12
Tour tooltips won't clear & elements not highlighted


adelavega:
If you don't clear the tooltip from the tour, it will hang around for a few more seconds even if you advance tabs. 

Also, it seems that after the refactor, the frontpage tour elements are not being highlighted.

PR #622, opened 2019-07-12
only show tooltips if on correct tab

rwblair:


adelavega:
Ok cool, the logout thing is working, but I'm still getting multiple tooltips if I go really fast.

rwblair:
fixes #623

rwblair:
fixes #620

PR #631, opened 2019-08-01
Predictor upload components

rwblair:
Base components for selecting a dataset, selecting a variety of files to upload, adding descriptions for predictor names from uploaded file present. Things on the todo list:
- [x] Post forms, run ids, and predictor descriptions to api.
- [x] Add some way to observe upload progress
- [x] Component to display users existing runs
- [x] Function to handle clicking on tab titles in predictor upload form.
- [x] Have cancel function clear form state for predictor upload form.
- [x] Add link to /mycollections route in user menu
- [x] Render tests for added components


adelavega:
I got this error compiling:

```

/neuroscout/neuroscout/frontend/src/predictor_collection/AddPredictorsForm.tsx
(2,18): Unused declarations are forbidden

```

adelavega:
Well only with `yarn start`. Compiled with `yarn build`

adelavega:
Weird, I'm also getting a CORS error when I switch the host to alpha. Will try locally.

adelavega:
Overall it's looking pretty good! Here's some minor feedback

Screenshot from 2019-08-22 18-38-02

- The buttons and lines from the box are quite close. A tiny bit of space would make things look cleaner.
- I think it would be better if the file picker for the first file was open by default, rather than having to click "Add" even for the first file.
We'll also need more instruction here, maybe on the first page, as its a bit bewildering without. Maybe at the top of each "Tab" we can have a bit of explanatory text.
- "Add more" rename to -> "Add event file". 
- I think it would be better if it said "OK" instead of "Hide". Also, it would be better if that hid the whole section including the file picker, and instead just showed the name of the file and the Edit button (Let's call it just "Edit")
- Although there are "Prev" and "Next" buttons, clicking on the tab doesn't navigate. That's probably OK but it has a hand icon when you mouse over, so its a bit confusing. 
- Also, about the buttons, I think having the "Cancel" button the Left makes more sense. Or potentially we could get rid of it, since there's an "X" at the top. And instead have the "Back"/"Next" button where it is. This would be a bit more like the Analysis builder.
- No colon ":" after "Collection Name" to match the style of the rest of the app. Same goes for the whole app in general. 

Aside from those tweaks, it looks pretty good to me!

In terms of future changes, we might want to let users click on existing collections and get more info about the variable names etc... but we can get to that later (when we discuss permissions, and whether these are live or not). I'll show this to Tal tomorrow if I get a chance.

adelavega:
- [ ] Private/public - default private. So that user collection are not shown to everybody. It's OK if we can only enable this on the backend. 
- [ ] Only show private predictors if they match user ID. 
- [ ] Add private/public status display to "My Collections"
- [ ] Add documentation

adelavega:
If people want to share, we can add a button to contact us.

adelavega:
- [ ] Add filter so that GET predictors only returns public preditors
- [ ] Add users private predictors to GET predictors call (not as urgent as we don't have users right now)

adelavega:
Having some compile problems:
```
Failed to compile.

/neuroscout/neuroscout/frontend/src/predictor_collection/__tests__/PredictorDescriptionForm.test.tsx
(8,25): Property 'testPredictor' does not exist on type 'typeof import("/neuroscout/neuroscout/frontend/src/predictor_collection/__tests__/data")'.


error Command failed with exit code 1.
```

rwblair:
latest 3 commits fixes the inactive datasets in predictor collection form and the google signle sign on avatar not showing on refresh.

rwblair:
@adelavega what text did you want for the "no data" in the predictor colleciton list?

I put in a yarn build in place of frontend tests until I get ssh enabled on travis to sort out the yarn tests.

adelavega:
"No custom predictors uploaded"

That should be good enough for now. Since so few people will look at this, we can add more info when we work on the docs.

Issue #624, opened 2019-07-15
Add system requirements


adelavega:
To run page. E.g. at least 8GB, etc...

PR #632, opened 2019-08-01
Add optional description to predictor upload

adelavega:
You can now pass an object/dictionary as `descriptions` when you `POST` to /api/predictors/collection`.

This lets you set the description for each of the columns in your events. In the future, we can expand this to add more meta-data but for now I think it's all we need.


adelavega:
Closes #624 as well

PR #634, opened 2019-08-04
Use json string for collection description

adelavega:
Merged #632 too soon. requests can pass dictionary, so i'm going to try dumping as string, and reading that in on the server side.


adelavega:
Great, looks like that works, will merge when tests pass.

PR #640, opened 2019-08-20
Index / db maintenance

adelavega:
There is an issue where the DB on disk is taking up 50GB. 49 of these are in the `predictor_event` tale. 

This is largely due to lack of compression / indexes, as the PG_DUMP is only ~600MB.

Well, this is also largely due to the poor decision to write the value for every predictor/onset in that table, even when the same is shown across subjects.

Here's what I get from running some stats on both `extracted_event` and `predictor_event`;

Extracted Event
image

Predictor Event
image

This shows that from the core relation (core data stored), `predictor_event` is only about 13x larger, which makes sense given the data duplication that sometimes occurs. 

The issue seems to be indexes. I got a little index happy when we had some performance issues. 

So I ran some stats on the indexes in the db:
image

It looks like the issue is that I have too many indexes on `predictor_event`, which maybe aren't even being used. 

At least it looks like I can safely drop `ix_predictor_event_run_id`

However, I'm not sure what to do about the unique_constraint `predictor_event_onset_run_id_predictor_id_object_id_key`. According to stats, this is not being used, however, it's a nice thing to enforce. However, its massive, so I'm thinking dropping it (and hoping we don't stick duplicates in), is a good idea. 

I also want to run `REINDEX` and `VACUUM(FULL)`, but to do so I need more free space than I have... 

I'm going to see if dropping one of those indexes is enough to at least let me do that. 

Thoughts? @tyarkoni @rwblair

### DIFF
--- a/neuroscout/models/predictor.py
+++ b/neuroscout/models/predictor.py
@@ -27,9 +27,6 @@ class Predictor(db.Model):
 
 class PredictorEvent(db.Model):
     """ An event within a Predictor. Onset is relative to run. """
-    __table_args__ = (
-        db.UniqueConstraint('onset', 'run_id', 'predictor_id', 'object_id'),
-    )
     id = db.Column(db.Integer, primary_key=True)
 
     onset = db.Column(db.Float, nullable=False)
--- a/neuroscout/models/predictor.py
+++ b/neuroscout/models/predictor.py
@@ -29,7 +29,6 @@ class PredictorEvent(db.Model):
     """ An event within a Predictor. Onset is relative to run. """
     __table_args__ = (
         db.UniqueConstraint('onset', 'run_id', 'predictor_id', 'object_id'),
-        db.Index('ix_predictor_id_run_id', "predictor_id", "run_id")
     )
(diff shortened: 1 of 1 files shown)


tyarkoni:
That all sounds reasonable to me. I do think it probably makes sense to eliminate the redundancy in `predictor_event`. I imagine it will be a pain in the ass to do, but I wouldn't be surprised if in addition to saving space it also speeds queries up some.

adelavega:
Yeah, it's been on my long term to do. Actually though it probably will increase query time, which is part of the reason we have it. In some ways, its sort of an "index", because we don't have to rebuild the structure of the stimulus time -> run time relation every time. 

I see what you mean though, it will result in a shorter table which will speed up direct queries though.

The real issue this is a problem is I'm running out of hard disk space. We could move the db to the other drive, but it'll be slower. Windows is still hogging ~100+ GB which is another outstanding issue.

adelavega:
Okay, that seemed to work. VACUUM(FULL) is a good idea to do regularly...

image

PR #648, opened 2019-08-28
always add subject level to bidsmodel #641

### LINKED ISSUE #641: Always specify subject level in BIDS Model
Need to check if this is true, but currently, if there is only 1 run per subject, the BIDS model doesnt specify a subject level (even its just a copy), and the first level gets passed to the group level.

Let's make it so subject level is always written. This is more explicit so you can always find the subject level maps. 

Also need to check if this is how I construct models with `pyns`

rwblair:
This will be blocked until fitlins is updated.

Closes #641


adelavega:
See #686

PR #649, opened 2019-08-29
add private boolean to predictor collections backend.

rwblair:


rwblair:
@adelavega could you generate migration for this on server? My migrations are in state that might cause problems if I pushed them to repo.

rwblair:
Would there be any issue with putting a FK from to predictor_collection directly in the predictor model? Do we ever for see a predictor belonging to multiple collection?

adelavega:
I should probably also implement the logic for only returing relevant predictors on a user by user basis.

adelavega:
OK, I merged to master the migration. I think the only thing needed now is the Schema change

adelavega:
Closing this in favor of #631

PR #650, opened 2019-08-29
Support uploading single subject files to NV

### LINKED ISSUE #630: Add more meta-data to NV uploads
Minimally:
 - Link back to NS
 - Set name to analysis name
 - DOI of original dataset

Are custom fields available?

adelavega:
Added support for single subject files in NV, and additional meta-data. Closes #630 

Also, to make the API more robust to connection failures, files are transmitted one by one in each POST request. This is tested and works.

To-do:
- Update `neuroscout-cli` to use new pyNS API, and find subject level files (+ effect and variance maps).
- NV needs to be updated to accept variance maps first 
- pyNS needs tests w/ single subject files.
- NV is not displaying all the meta-data we are now sending over (link back to Neuroscout).
It does seem to display the `full_dataset_url` link (which is now going to the actual dataset). We could use that as a link back to neuroscout if we want an easy fix.


adelavega:
Working w/ new version of pyNS!

PR #654, opened 2019-09-03
Remove PredictorEvents from Predictors derived from ExtractedFeatures

### LINKED ISSUE #642: Redundant store / processing of predictor_events
Two problems:
1) Database size (10-50GB for predictor event table)
2) Query / Report / Compilation speed
 - Mostly compilation of the identical runs across subject
 - Querying itself I think is not so slow

Solutions:

Neuroscout
- Add an intermediate representation of "unique" runs ("configurations"). This would be upon ingestion, but would map stimuli to these proto-runs, which are them mapped to real runs.
- Skip adding this intermediate representation and see how slow it is to re-create the current API without the `predictor_event` table. So essentially, ignore this table, and try to recreate all functionality
- If the repetition units are known, or can be computed, only ask pybids to compile certain runs

Pybids
- API for Sparse Run Variable
Instead of writing out files to speed things up 
- More elegant build in version of detecting repetitions

adelavega:
Fixes #642 

It's easy enough (if a bit redundant) to dynamically generate `PredictorEvents` for `Predictors` from extractors. It remains to be seem how slow this is (especially for Reports).

However, we still need the `PredictorEvent` table for `Predictors` that come from event files (and could differ between subjects), and uploaded PredictorCollection (as they don't link to stimuli). We could potentially add another table here to reduce redundancy but it doesn't seem worth it.

### DIFF
--- a/neuroscout/core.py
+++ b/neuroscout/core.py
@@ -48,7 +48,6 @@ route_factory(
         ('RunResource', 'runs/<int:run_id>'),
         ('PredictorListResource', 'predictors'),
         ('PredictorResource', 'predictors/<int:predictor_id>'),
-        ('PredictorEventListResource', 'predictor-events'),
         ('PredictorCollectionResource', 'predictors/collection'),
         ('UserRootResource', 'user'),
         ('UserTriggerResetResource', 'user/reset_password'),
--- a/neuroscout/populate/extract.py
+++ b/neuroscout/populate/extract.py
@@ -17,7 +17,7 @@ from pliers.transformers import get_transformer
 from pliers.extractors import merge_results
 
 from ..models import (
-    Dataset, Task, Predictor, PredictorEvent, PredictorRun, Run, Stimulus,
+    Dataset, Task, Predictor, PredictorRun, Run, Stimulus,
     RunStimulus, ExtractedFeature, ExtractedEvent)
 from .annotate import FeatureSerializer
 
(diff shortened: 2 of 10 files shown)


adelavega:
Now need to test this on master... then if all tests are passing:
1) Back up DB
2) Delete PEs derived from EFs

adelavega:
Tested on Master, and seems to work. Backed up DB.

adelavega:
@tyarkoni this cuts down the `PredictorEvent` table by about 75% (currently at 10GB). We still need that table for events that are not associated with an `ExtractedFeature`.

The main potential downside here is this could slow down Reports (to dynamically re-create the timing), but it actually seems like a wash in my limited testing.

To actually speed that up (which IMO is acceptable in speed right now), I think we need improvements with pybids. To speed up compilation, we could skip redundant validation (i.e. across subjects that saw the same stimuli).

adelavega:
Again, to be clear, this only changes the database representation, and shifts the creation of `PredictorEvents` to report/compilation time to save db space. We are still writing out event files for every run.

We are not detecting if events are similar across all runs (and could hence only write on event file).
This seems potentially hard though because if events that vary across subjects (e.g. RT) are included in the model, we need to account for that. So I think I'm okay with leaving it as is.

tyarkoni:
Haven't reviewed the code, but that all sounds good to me. I agree that optimizing PyBIDS is a sensible next step (though it's not going to be a high priority).

adelavega:
Cool. I'm going to test w/ another model (from start to NV upload) to ensure its working. If all is good, I'll merge and purge those rows from the db.

adelavega:
Without building the pybids object, compiling took ~17 seconds, and 9 minutes doing `analysis.setup()` in pybids.

This can be disabled in the frontend, so its an option to speed the process up.

Otherwise, all looks good. Merging.

Issue #599, opened 2019-06-18
Move domains to www.


adelavega:
Tal has already redirected neuroscout.org www.  to alpha.neuroscout.org and i've renewed the https cets.

However, ideally, the main site would be at www. (and it would be labeled as beta).

This should be easy, but I want to coordinate the switch, to minimize any downtime. 

I think the thing to do will be to redirect neurscout.org and alpha to www., and set that as the primary domain in nginx, and the various internal variables references in the back and frontend.

adelavega:
Closed

PR #658, opened 2019-09-10
Migrate domain to bare neuroscout.org and add redirect

adelavega:
Adds redirect from `alpha.neuroscout.org` to `neuroscout.org` and changes the official domain to the latter.

@tyarkoni can you check that the DNS for `www.neuroscout.org` is set up correctly? I was able to extend the cert to include the bare domain and alpha, but www. does not work. (Now I'm timed out from trying again for a while)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,7 +47,7 @@ services:
   swagger-ui:
     image: swaggerapi/swagger-ui
     environment:
-      - "API_URL=https://alpha.neuroscout.org/swagger/"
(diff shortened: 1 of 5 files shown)


adelavega:
Closes #599

tyarkoni:
Sorry, `www` was pointing somewhere else. Should work now (or in a few minutes, once DNS propagates).

adelavega:
Okay, this all seems to be working. www is acting funny with https but that seems like an independent issue that might have to do w/ caching.

Issue #666, opened 2019-10-02
Incorrect analyses showing when logging in and logging out and logging back in.


rwblair:
force reload analyses on login?

PR #667, opened 2019-10-02
Fix analyses not showing on logout -> login

rwblair:
Also clears analyses on logout. Also fixes a problem with checking the truthiness of isGAuth in the authstore.


adelavega:
Fixes #666

PR #669, opened 2019-10-03
update neurovault upload status component/api calls

### LINKED ISSUE #652: Update frontend to look at NV file status
In #650 I added support for file-by-file uploads (and single subject files), and as a consequence the API changed.

Now there is a `files` attribute which lists the individual status of each file being uploaded.

@rwblair , can we update the frontend to reflect this? One minor downside is this breaks compatability w/ previously done uploads. Maybe I can change the API to summarize the status across files (e.g. "PENDING" if any files are pending, and show a list of errors).

rwblair:
fixes #652


rwblair:
There is an assumption right now that there will always be a collection_id returned from the api.

adelavega:
Another minor comment: Instead of putting the box header to be the URL, make it "Collection ID: {id}" with the link staying the same

PR #670, opened 2019-10-09
predictor table header width fix

### LINKED ISSUE #618: Misalignment of column headers and row width in Predictor table
image

rwblair:
fixes #618


adelavega:
very awesome, looks good thanks

Issue #540, opened 2019-04-15
Add ReplaceNA option to Scale


adelavega:
There are a few missing options for transformations.

Most saliently, for `Scale` we should have `ReplaceNA`

adelavega:
This requires adding a "Drop down" select menu as the three options are `None` `before` and `after` (not a boolean, as other fields)

adelavega:
Closed with #671

PR #671, opened 2019-10-09
Add  ReplaceNA to scale xform, fix unselect all predictors error.

### LINKED ISSUE #590: t.filter is not a function error when unselect all runs
This happens only once when unselect all subjects

rwblair:
fixes #590, #540


adelavega:
Cool, LGTM

PR #672, opened 2019-10-10
Add sentry to front end and backend.

rwblair:


adelavega:
I wonder if we should add sentry to the celery worker. Seems like a good idea actually.

PR #675, opened 2019-10-14
Update Sherlock Merlin

adelavega:
Re-enable and update the Sherlock Merlin dataset.

- Fixes bug related to pybids > 0.8. `layout.get_collections` would return incorrect result because preproc and raw file interfered in variable loading.
- Ensure re ingesting does not cause conflict with old `RunStimulus`


adelavega:
TODO:

- Catch up Sherlock Merlin on extracted variables

PR #703, opened 2019-12-11
small fixes to docs

rbroc:
fixed typo in dataset description and added note on absolute paths needed to run analyses via neuroscout-cli.


rbroc:
More potential edits:
- Add more details on sparse/dense feature representations?
- Add more details on how data are preprocessed
- Add overview of which features are available for each dataset/modality
- Add more details on na replacement during transformation (why/when to replace NAs before rather then after when rescaling)
- Potentially more info on terminology ("task", "run", "condition")
- More info on stats behind model estimation (e.g. the concept of "fixed effects meta-analysis" at the subject level? Maybe this details can be included in the paper though)
- Potentially add details on relevant Docker commands and flags in Run analysis section

PR #712, opened 2019-12-20
force reconstruct of review tab for fresh results on model change

rwblair:
This is inefficient, every time the review tab is navigated to it hits the api to redownload results.

Before I had the conditional around the reports tab in builder the report state would properly update with the new reports as necessary but the default active plot in the plots tabs would not rerender with the new results.


adelavega:
LGTM, the lag is not really noticable. If it becomes an issue we can store the old state somewhere else so it doesn't have to do an HTTP request. This might be more of an issue with slow connections but I think it's fine for now.

Issue #686, opened 2019-11-20
Update model spec for pybids 0.10


adelavega:
Not released yet but need to update `DummyContrasts` and `FEMA`

Issue #697, opened 2019-12-11
exporting and inspecting design matrices in review tab


rbroc:
Possibly add the option to export design matrices from the review tab for further inspection.</br>
Would also be great to be able to navigate design matrices for different runs and subjects.

adelavega:
Fixed with #716

PR #716, opened 2020-01-09
Add ability to generate and view reports for any run.

### LINKED ISSUE #533: Allow user to select subject/run to display design matrix for
I could have sworn there was an open issue for this, but maybe we only discussed it in person. Anyway, on the review page, it would be nice to let users choose which run/subject they want to display the design matrix for (and have it generate in real time if not already cached).

rwblair:
Fixes #533 #686 

Currently no warning on number of runs selected. @adelavega any idea at what number we might want to warn at?

This also assumes that the api returns run plots in the same order that front end asks for them.

### DIFF
--- a/neuroscout/frontend/src/analysis_builder/Builder.tsx
+++ b/neuroscout/frontend/src/analysis_builder/Builder.tsx
@@ -1066,6 +1066,7 @@ export default class AnalysisBuilder extends Reflux.Component<any, BuilderProps 
       unsavedChanges
     } = this.state;
 
+    let reportRuns = this.state.availableRuns.filter(x => this.state.analysis.runIds.find(runId => runId === x.id));
     let isDraft = (analysis.status === 'DRAFT');
     let isFailed = (analysis.status === 'FAILED');
     let isEditable = editableStatus.includes(analysis.status);
@@ -1223,7 +1224,7 @@ export default class AnalysisBuilder extends Reflux.Component<any, BuilderProps 
                     <div>
                       <Report
                         analysisId={analysis.analysisId}
-                        runIds={analysis.runIds}
+                        runs={reportRuns}
                         postReports={this.state.activeTab === ('review' as TabName)}
                         defaultVisible={this.state.doTooltip && this.state.activeTab === ('review' as TabName)}
                       />
--- a/neuroscout/frontend/src/analysis_builder/Builder.tsx
+++ b/neuroscout/frontend/src/analysis_builder/Builder.tsx
@@ -41,7 +41,7 @@ import {
   TransformName,
   TabName
 } from '../coretypes';
-import { displayError, jwtFetch, timeout, isDefined } from '../utils';
+import { displayError, jwtFetch, sortSub, timeout, isDefined } from '../utils';
 import { MainCol, Space } from '../HelperComponents';
 import { config } from '../config';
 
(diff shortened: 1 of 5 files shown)


adelavega:
Here's a fun state related bug I found. When I go to an already compiled analysis, it lands on the Review tab. This tab is blank when you first land on it.

Try it here: https://neuroscout.org/builder/MNW3V

But if I go to "Status" then back to "Review" it loads properly...

rbroc:
Format looks awesome, just a couple of points: 

- You can type stuff in the search box but it does not seem to be searching on the basis of the same type of subject IDs used in the dropdown list (if you try typing "s" in the searchbox, it would not match any IDs, despite all IDs in the dropdown list being "subject-something" - e.g. in AxkxB).  Could we switch to searching based on same ID type used in the dropdown? 

- testing this with analysis AxkxB (compiled and public). The report page first loads "subject - 15". If I add one more subject to that, it displays two tabs but in both tabs there's the same design matrix (note that I am using fmriprep motion parameters as regressors, so they *should* be different). Maybe I am missing something, but see if you can replicate that.

- Agree on the ordering of subjects in the dropdown, it's a minor thing but it'd make it nicer. 

- Being able to inspect different runs would be awesome!

rwblair:
Latest push should address 
- hanging tab on run removal from select bar
- run ordering in select list
- allow filtering in select 
- Only runids selected in overview should be present in select list.
- Review component not displaying on initial load of passed analysis

adelavega:
Awesome, just tested your latest commit and I think almost everything is addressed (including showing all the runs, and filtering based on the runs selected in the first tab).

I think the only remaining concerns are:

- Move into a modal? I think we can probably wait on this, and do it in another PR if it keeps bugging us.
- Warn if more than 3 runs are selected at once. I think this might be more impotant. Maybe once a 4th is added, a pop up warning says something? Or it could be on the "Generate Report" click that it makes you confirm yes/no if you really want to do this. 

Otherwise, LGTM!

adelavega:
Hey @rwblair can we also take care of #697 here? It would simply be a link to `design_matrix` which is given by the report. I think a button below the design matrix plot would suffice.

rwblair:
@adelavega I've added a download link for the correlation matrix. There is a download link in the vega embed but it isn't clear that there is even a menu until you hover over it, much less what its for.

Also right now I'm generating the PNG when the correlation matrix component is mounting. We can push this off to until the button is clicked if its a performance issue but it will take some work. (think we need to generate png from event, create an anchor tag, add it to dom, and the use its reference to call its click() function to have the browser's download dialog pop up)

rwblair:
Added link to download design matrix as a png. This png is being done when the design matrix component is loaded. We can push this generation off to when the download link is clicked if this is a performance issue but it will take some work ( think we need to generate png, make anchor tag, add it to dom, use reference to anchor to call its click() function.)

adelavega:
Oh, sorry Ross. Miscommunication, I meant a link to download the .tsv design matrix. I think you can already download thos einteractive plots by hovering over the `...`

adelavega:
The tsv is available as the key `design_matrix` in the `result`. The way the link looks would be fine. I also wonder if the last change introduced a bug, because now I'm requesting two runs, and getting back only a single tab. Hopefully reverting these changes will undo that.

rwblair:
Ah good I'll revert the changes. Knew about the menu in vega embed, but figured we might want it more explicit? Any way downloading tsv much better.

adelavega:
On a very wide screen, like my external monitor, the link gets split up:

image

adelavega:
Otherwise, LGTM!

rwblair:
Try that latest fix commit, I tried zooming out on my low res monitor and couldn't get it to float to the side.

adelavega:
That fixed it, merging!

PR #723, opened 2020-01-16
Dedupe auto contrasts

### LINKED ISSUE #678: Default contrasts should never generate two contrasts with same name
- In addition, the list should be updated if the inputs change
- Maybe just have a check box for default contrasts

rwblair:
Fixes #678

Also when a predictor is removed from the predictors tab any contrast that includes it is also removed.


adelavega:
Okay, looks good, but actually while we're at it, can you remove any transformations that apply to removed Predictors?

rwblair:
Ok added!

adelavega:
LGTM

Issue #707, opened 2019-12-17
Disable next button (on dataset selection) is task is unselected


adelavega:
This is an issue for when two tasks are available, and we go back and change datasets. The next button is still enabled, and then no predictors load.

Issue #692, opened 2019-12-09
Selecting all runs ignores task


adelavega:
When you select a task (e.g. `MerlinMovie`), only 18/36 rus are selected (because the other half of runs belong to the other task).

But when you check the top "all" checkbox in the run selector, *all* runs are selected including those for the other task. 

This is pretty minor, but its worth nothing.

We discussed removing the run selector (or making it only for advanced use), and instead having selectable entities (i.e. select task, run_id, subject_ids), which might makes this not a big problem at all.

adelavega:
This also happens if you click "Clear Filters"

adelavega:
OTOH this could be seen as an advantage, as you could analyze the whole dataset, across both movie in one go. 

We need to think about this though because currently I ingested piecemeal, so the two tasks have different Predictor ids for the same predictor. We could be deliberate about this and ingest at once for all tasks instead.

adelavega:
Okay, let's only allow one task at a time, and lets filter the run table inputs by the selection in task

adelavega:
Bug is only text that says "All selected" should instead say "18/36" for example.

adelavega:
Closed by #689

Issue #717, opened 2020-01-10
NV file uploads with more than 100 chars fails


adelavega:
Single subject files, which tend to have filenames longer than 100 chars, will fail on NV right now. 

The long term idea is to migrate neurovault to accept filenames longer than 100 chars. In the meantime, may shorten filename, and/or disable single subject file uploads (as these are a concern in any case).

adelavega:
https://github.com/NeuroVault/NeuroVault/issues/726

adelavega:
Not entirely taken care of by #724, but now it's much less of a concern. Closing since its a NV problem.

adelavega:
This is coming up again due to sensorimotor norms. I think the sensible thing to do is truncate file names longer than ~86 characters in the base name to be safe

adelavega:
Maybe also remove the `_stat-t` as this is encoded in the `Type` column

adelavega:
Closed by NeuroVaul migraton!

PR #724, opened 2020-01-16
Dataset list refactor

### LINKED ISSUE #689: displaying more dataset info in analysis builder
It might help to be able to see more details on the available datasets when you have to select them in neuroscout.org analysis builder. </br> Could be useful to expand the **datasets table** displayed in the **analysis builder** , potentially adding: 
- more systematic info on which modalities are available in the dataset (e.g. a yes/no value coding for presence of "auditory", "visual", "speech" info)
- duration
- TR and/or number of TRs per subject
- tasks (passive viewing/listening or other)
- dataset id used to retrieve dataset details when accessing API via pyNS
- when dataset is a subset of the linked dataset, pointer to how and why this has been done 
- reference to most relevant paper?

adelavega:
Closes #689 

Let's also take care of #707  and #692  here.

For now just made these small backend changes:

- `mimetype` for datasets are now sorted by id, so that the oldest stimuli for each dataset goes first. Let's display the mimetype for only that one, as that is the original stimulus.
- `TR` is now shown for tasks under datasets route.


adelavega:
We could also add a `long_description` for each dataset, or `task`. Probably better for `task`.

adelavega:
Author list sometimes was one long string. I manually updated these entries, as these are technically BIDS errors which need to be corrected prior to ingestion. In the future, check this field prior to ingestion.

rwblair:
Length of stimuli for display.

adelavega:
@rwblair fyi the server is live with the changes in this branch

adelavega:
The API, I mean.

rwblair:
I didn't end up putting any of the task information in the dataset expansion, just extended the task table to include it.

No real formatting going on in expansion still, will only show mean age, percent female, and known issues if present.

adelavega:
I actually think the Task information might be good where it is now that I've looked at it.

My main gripes now will be aesthetic ones, but I may play around with it myself and see if I can think of other ideas

adelavega:
- For "Average Run Length" let's just display `s` instead of `seconds`.
- Remove the (N) after subject

adelavega:
I think I'm going to change `description` to `summary` and shorten it, then add a longer description for each dataset. This will include `known_issues` so I will get rid of that field.  The `description` will be displayed under the + button, and the `summary` where `description` is now.

I think for missing demographics we should put `n/a`, so that it doesn't look too empty. We could potentially put these in a box that says demographics.

Let's also put a complete author listing under the + button (or maybe what we were doing before that showed ~5 others, with `...` in between).

Let's preface the external link with something like `References and Links:`. 

Curation wise, I'll do the following:
- Put all author lists under the same format
- Change the descriptions/summaries
- Add long descriptions for all.
- Add missing demographics for HBN, other tasks...


Other ideas for other fields (that as in OpenNeuro for example)
- Date added (not sure if I even have that)
- Dataset DOI
- Funding

adelavega:
Snuck in a patch for #717

adelavega:
Okay @rwblair I think the fixed width looks good!

@rbroc will you take look at the final descriptions and propose changes as you see fit?

adelavega:
@rwblair one minor thing here

Looks like when I unselect all runs, then select all again (in SherlockMerlin, with a task selected), the following happens:

image

adelavega:
Otherwise, eveything else LGTM,

rbroc:
> Okay @rwblair I think the fixed width looks good!
> 
> @rbroc will you take look at the final descriptions and propose changes as you see fit?

Looks great!
Really minor things if one wants to be super picky:
- For HBN, maybe woudn't say that only a "small" subset of participants is available (we have 142 subjects after all)
- For schematic narratives, do we also have the data from the audio-only runs? If not, I would make that clear in the short description. If yes, would fix "description" in task tab, which right now says "audiovisual watching".
- Typo in Forrest, should be "movie being present**ed** in German", I guess.

Everything else looking awesome!

adelavega:
@rbroc I think in the Schematic Narratives, the audio only clips are played during the video stimulation within a run, so we're good.

Issue #704, opened 2019-12-12
Throw error when transformation is not applicable


adelavega:
This is more of a pybids issue, but when you try to scale a variable with all 1s (speech), it fails, and silently drops the predictor.

Instead, lets throw a warning/error in the reports tab

adelavega:
Closed by #725

PR #725, opened 2020-01-16
Generate warnings in report

### LINKED ISSUE #441: Warning messages in reports
Currently, report failures just show an ugly traceback.

It would be nicer and more informative if we were to catch certain types of errors (that are common) and gracefully fail, and offer suggestions. 

For example: if `n/a`s are found after transformations are applied (and thus would crash on fitlins), give a warning and suggest dropping n/as or scale + replace n/a.

adelavega:
Closes #441 #704 

Adds a `warnings` field to the `Report` model. This field is update if any warnings are generated, and even if the Report ends up crashing. 

We should display these above the traceback using ant's alerts, such as this one:
Screenshot from 2020-01-16 17-02-31

If each warning is pretty short, we should be able to stack them above the traceback (if any).

Sometimes warnings will be spit out even when there are no errors and the design matrix is compiled.

Maybe they can go after `Correlation Matrix`? These are not going to be subject specific so they should not be included within a Run tab.

I will next add a warning for #704 to test this out.


adelavega:
Cool, it's working on the backend

adelavega:
Alright, all is looking good. @rwblair please take a look next week and test & see if there's a cleaner way to do things.

On the backend, still need to catch errors in the warnings module. Uncaught errors in Reports will result in Report hanging indefinitely.

Otherwise, just need to add more warnings. For now, focus on checking if `n/a`s are handled correctly.

adelavega:
- Try to generalize as many warning as possible into a `Validator` class in fitlins that checks all internal arguments that starts with `_check`

- Add wrapper that points to Neurostars / Github issues

PR #734, opened 2020-01-31
Context-dependent word level estimates

### LINKED ISSUE #713: ComplexText Stimulus Representation
A general problem we haven't solved yet is how to tokenize force aligned stimuli to take into account context.

Currently, all extracted events (EEs) are generated from a single stimulus a time.

For example, when we compute the frequency of a word, we do it on a word-by-word basis (1-to-1 mapping between stimulus and EE).

We also have situations where a single stimulus results in a multiple EEs. For example, `shot_change` is applied to a single video file, but results in multiple EEs (1-to-many).

For speech, we would like to be able to get features at the temporal resolution of the force aligned text, but that take into account the context occurring prior and after the word. For example, if there was a pliers extractor that computes the number of times a word has occurred in the movie so far (this would be a many-to-many mapping).

For a similar situation (face_net, `first_time_face`), we solved this by operating entirely outside of pliers/neuroscout and just uploading a custom `PredictorCollection`. 

I think the solution here is to have a `ComplexText` type of Stimulus representation that groups together individual words, with individual onsets, but has a single RunStimulus association back to Predictor time. 

This would require an additional table in the database that keeps track of the individual words and their onsets, and maps them to a single `Stimulus` entry of type `ComplexText`.

Alternatively, we could save these "tokenizations" as .tsv file which are read in by pliers as `ComplexTextStimuli`, but I think I would rather represent these as a new table in the db.

This should allow us to tokenize all the speech in a run, or even a while movie across runs, and then map that back into Predictor / original "functional" time like we do with any other stimuli. 

We could need to create some functions to create these "tokenized"/"grouped" ComplexText Stimuli from individual words with onsets in the db.

adelavega:
Closes #713

My preferred solution is one that is slightly hacky but fits in with the current way of doing things as much as possible.

When ingesting transcripts, each word is represented in an abstract sense, with no timing information. Here, I'm adding a representation of a whole transcript, with `content` including the whole text for a run. These are differentiated by a mimetype of `text/csv` (to indicate multiple rows) instead of `text/plain`

On runtime, these will get recreated into `ComplexTextStimuli` with word level onsets (by querying the db). 

Then on extraction, extractors that require the whole transcript will have to be denoted in a different way. I'm thinking this may be worth having its own field in the `JSON` object.

In terms of tokenization, we also need to specific ways to then split this `ComplexTextObject` for approaches such as sliding window, or even for those that require the whole text. 

Still need to:

- [x] Recreate `ComplexTextStimuli` from db for existing datasets/transcripts
- [x] Add custom extraction logic for these objects (recreate CTS object from db)
- [x] Implement tokenization/moving window logic.


adelavega:
@tyarkoni (cc @rbroc)  I notice BERT returns a list, in a single field, rather than `n` encoding columns. 

Currently, we store all values as strings in the db, so we have to be careful about how we handle this. Would splat-ing this array onto individual columns by pybids' job, or custom neuroscout logic prior to ingestion to ingest as separate features? We could also allow a `JSONB` field to allow arrays of values, which would get split into individual columns at report/compile time. However, this would make it incompatible with the frontend for the most part, as the column names would be unknown ahead of time. 

Otherwise, I've tested the BERT extractor, and the windowing logic is implemented, so once we figure that out we should be ready to go.

tyarkoni:
I see this as NeuroScout's job to handle. @rbroc and I talked it over, and from a pliers standpoint, I don't like representing each dimension as its own feature, because the user will almost always want to use that vector as a single variable. So I think probably you'll need to break this up during ingestion. (I agree that passing it to PyBIDS to deal with doesn't make sense.)

tyarkoni:
I suppose we could consider adding an initialization argument to the pliers extractor that controls whether the embedding is returned in one field or one per dimension. I'd be fine with that if it turns out to be a PITA to implement the custom logic in NeuroScout (though it doesn't seem unreasonable to me to think this is something we might eventually want anyway, for other features). If it's not too much work, I think I'd prefer sticking with doing it in NeuroScout.

I wonder if we could just adopt a universal heuristic that *all* iterable values found in pliers features are to be broken up and serially numbered during ingestion into NeuroScout. Are there cases where that policy would fail us?

adelavega:
I think your last suggestion probably works.If value is list, create `n` number of features that end with `_{n}`. 

The main downside is if we ever want to include these features on the frontend, it will be unwieldy. In that case, we'd need a way to group features.

adelavega:
Also, it just seems that you probably want to use less than all the embeddings, so having them as individual features is less useful. That said, that's what they literally are so its consistent with the way we do things.

tyarkoni:
I think we'll probably want to include all of the embedding dimensions in most models where any are included, so I agree that it would be good to have some front-end short-cut for grabbing/grouping all of them (while also allowing the user to select individual ones). But I guess this isn't really any different from how we handle confounds,  right (except for there being more dimensions)? Do we currently require users to check each confound box individually, or can they just "select all" to do it once?

adelavega:
There's no groupings for confounds yet. That said, it's fairly easy to do with the search, since there is a check all box for all results. 

Either way, sounds like grouping is something we could handle separately. For now, not to overwhelm the frontend, I'll set these Predictors to not "active". pyNS can still work with these, they just won't be display by default in the frontend.

adelavega:
Partly we don't have this because you may not necessarily want to select all. For some, the first 6 noise components (out of a variable number) may be sufficient.

adelavega:
I added the following to how `feature_schema.json` is processed by `FeatureSerializer`:
If there is a `name` or `description` attribute, you can not only use regex to plug in matching groups from the pattern match, but python string formatting variables (`{var}`) are used to indicate extractor attributes.

So you can do something like:
```
        "encoding_(.*)$": {
          "description": "Encoding for {pretrained_model}, windowed with {window_n} letters {window_method} target."
          "name": "{pretrained_model}_{window_method}_{window_n}_encoding_\\1",
        }
```

which would result in a feature name like: `bert-base-uncased_pre_25_encoding`, without any custom logic, aside from the feature schema. 
also, I added `splat` functionality, such that if a value is a list, `_{n}` is atttached to the feature name, and each value is its own feature. For features other than complex text ones, I've left this off, and set to throw and error so that we don't accidentally ingest tons of features. 

Finally, I added logic to `FeatureSerializer` so that optionally you can keep only the results from the value with `object_id` = `max(object_id)` for a given `ExtractorResult.to_df`.

adelavega:
After removing the `history` column, the total size of the `extracted_event` table is 76M rows at 636MB. 

A single extraction of a `PretrainedBertExtractor` creates around 1.1M rows which is about 93MB. 

This is about 1461 events per run, which less than one per second on average. This actually seems fairly reasonable. Obviously, the events will be higher frequency when speech is occuring, with stretches of no speech.

For now, will try: rounding values to 5th to see how this affects the size. 

Then run models with both and compare.

adelavega:
Looks like rounded and unrounded values identical. So much so that I think we can round even more if necessary.

adelavega:
Okay, looks like I'm using the right column type for `value` and rounding did shorten the column's length/size on disk from about 20 to 8-9 bytes. 

However, the problem is that each row has about ~72 bytes of overhead, including the onset, duration, object_id, foreign keys and necessary indexes (primary_key, foreign_keys). 

That means that assuming 90 bytes per row * 768 embedding dimensions * 1400 words = 96768000 bytes = 96.77 MB

Assume we do this for all runs with speech in Neuroscout (~25), that's about 2.5 GB. Not totally crazy but not very efficient. 

For reference, if we were to store all dimensions in one row as a string (which we then parse), it would only be about 164 MB for all datasets.

adelavega:
Another solution is to store values for predictors we know to be large and repetition as a `JSON` column in `Predictors` (still allowing us to split predictors for selection).

If stored as a list of list, we could use 25% of the size it would be to store as proper rows, and minimize the custom code to deal with this (only upon dumping to `PredictorEvents`).

adelavega:
Looks like everything went well with running the BERT analysis, so ready to merge.

PR #739, opened 2020-02-20
Extract Sensorimotor Norms

adelavega:


adelavega:
@rbroc These should be extracted for all datasets now!

PR #754, opened 2020-04-27
Fix logic when regenerating available runs from existing model.

### LINKED ISSUE #752: Report generation, even though compilation works
Sometimes, report generation fails through the web-ui, saying "incorrect run-id" specified.

For example see: https://neuroscout.org/builder/9BqzZ

This happens on valid designs

rwblair:
Fixes #752


adelavega:
Sweet LGTM thanks

PR #773, opened 2020-06-02
Fix division by zero error when empty predictor

adelavega:
Fixing upstream pybids issue with handling single event predictors (or those with all `n/a`s) by adding a duration of 1 to the empty events.

Fixes #773


adelavega:
This is a pybids issue when an event only has a single onset

PR #775, opened 2020-06-04
Speed up reports - optimize creation of PEs

### LINKED ISSUE #755: Some analyses are much slower than others generating reports
Reports are again agonizingly slow... maybe 300s sometimes for a simple model.

Need to look again as to how to speed this up and what the pinch points are

adelavega:
Closes #755 

Looks like my original code for creating PredictorEvents from ExtractedEvents was terrible inefficient. 
Changing it to do more in SQL looks like could speed up this part of report creation substantially.


adelavega:
Some truly embarrassing speed ups (my original function was *bad*), by redoing as much in SQL as possible, and adding indexes.

In an example report w/ 6 predictors, report took 6 seconds to complte, and db query only account for 500ms.

Remaining speed ups will have to come from pybids, but they seem acceptable for now.

tyarkoni:
Great!

PR #782, opened 2020-06-11
Add stimulus timing to /predictor-events route

adelavega:
This PR allow users to say `predictor_timing=True` in the `/predictor-events` route and get back information about the stimulus associated with the event, including the timing information (for that run) and path (truncated) of the stimulus used.

For now, I'm not serving the stimuli to the world, so this is only useful internally.


adelavega:
@rbroc can you let me know if you have any feedback?

this work well for you?

PR #783, opened 2020-06-15
Fix extractor source not updating in certain cases.

### LINKED ISSUE #708: Logic for filling in extractor source is unreliable
Currently, the `source` field is only populated by the frontend (from extractor name), when the length of `availablePredictors` changes. 

This is unreliable because there's a chance two sets of runs (i.e. tasks/datasets) will return the same length of `availablePredictors`, which will fail to trigger this logic.

rwblair:
memoize setting the extractor source, update only on predictor change. Use deep equal instead of array length to determine if this happened. Fixed another issue around setting selectedTaskId not working when loading an analysis from a dataset with multiple tasks.

fixes #708


adelavega:
cool. ready for review?

Issue #417, opened 2018-12-05
Show authors for public analysis listing


adelavega:
But what should be shown? Display name? email?
Email is too private, but display name needs to be set.

We need to force users to choose a display name (not user name). We could auto generate one based on their email, but allow them to edit it via "My Profile" (#159)

rwblair:
Fortunately we can get a nice display name from the google auth. Maybe we add a field in signup for a display/user name?

adelavega:
Closed with #791

Issue #159, opened 2017-10-18
Add user profile


adelavega:

adelavega:
University? Affiilation?

adelavega:
Instead of asking at the time of registration, we can have a user profile that can be filled out with such information. Low priority for now.

adelavega:
Display name, change password should all go here.

adelavega:
New fields to add to backend:
`university / company`
`display_name` if None, frontend displays `name`
`website`

In terms of frontend, all fields except `password` and `email`, can be edited and simply "saved" which does a `PUT` to `/user`

Email change:
This is allowed through the `PUT` route, but need to add to the backend a trigger to reset email verification email. I think it's better if its displayed separately on the frontend as this is a a bigger change. Maybe a warning can be shown saying the email will need to be verified.

Password change:
Require old password, verify by getting new JWT
Require user to type password in twice, verify they are the same. Requirements for now are  minimum 6 chars.

adelavega:
Maybe we should also have a `DELETE ACCOUNT` section?

adelavega:
Can we pull a user name from google?

adelavega:
Need to add `Insitution` and to sign up sheet, and route SSO users to that sign up sheet. 

Then add a `My Account` that lets you change your: `Name`, `Institution`, `Email`, `Password` and `Anonymous` status.

If someone sets their profile to `Anonymous` any public analyses shared by them will not show their name.

We can then add pages for users e.g.`neuroscout.org/users/{___}` although that seems less necessary.

One issue here is we don't have "user names" per se, and use the email as the main identified. We could potentially change and add a username for everyone (could just go off their google id or the first part of their email), or just go by the internal db id.

PR #791, opened 2020-06-29
Add user profiles

adelavega:
Merge Summary (8/14/2020):
- Added user profiles for each user under "My Profile". Displays public profile, with publicly viewable analyses only.
- Added `user_name` for all users. Set to default value based on name, but can be changed by user.
- All profile values can be edited by users under "My Profile"
- Authors are linked in public profile view to profiles
- Authors are also linked in analysis summary

_____________________________

WIP PR starting #159 

- Added `Institution` field to db

Frontend todo:
- Add `Institution` to sign up sheet form
- Re-route Google users to let them set their name and institution when signing up that way.
- Add a `My Account` page where users can edit their information
- Add public profiles for users that can be clicked under an analysis (#417). These profiles could summarize their profile info + number of analyses created (or even a listing of their public analyses)

Questions:
 - Should we an add an `Anonymous` privacy mode. If this is set to true in `My Account` then their profile would be private to others, and any analyes shared by them would not show their name. I think this is probably unnecessary as users can just keep their analyses private, and theres no listing of profiles
- Should we add a `username`? Currently we store `google_id` to this is a starting point. We could force people to use a username, which for now we could set to the first part of their email. This could be what is displayed instead of their name, and would be the URL for profile pages.


adelavega:
Should we let users upload a profile pic besides their google one?

adelavega:
Minor thing but I'm not sure why your update to the user schema broke the tests.

I'm not sure if the test that's failing makes that much sense anyways. Basically, the schema should reject partial updates, unless the mandatory fields are included. But it looks like now it's allowing a single field update.... not sure why that's changed but it looks OK

rwblair:
@adelavega last two commits have public profiles. Currently not linked anywhere. I have them using the user id from the database for both the front end and api routes. Do we want to identify them by some other method? Email is guaranteed to be unique but encoded in the url its a lot more busy looking than just an id. We had also talked about generating another type of id, maybe on like for analyses?

Fixed the clone being visible for non logged in users.

Also right now it only displays fields for which there are values in the user profile, not sure if we want to default to just showing empty entries.

Finally I still need to make the twitter handle and personal site and actual link.

adelavega:
Looking pretty good.

We should probably add a link in the profile editor to see public profile, or have the profile link go to that first, and then if your viewing your own profile, have there be a link to the profile editor (with a link back probably as well).

I'd be nice to see a bit more visual structure, like maybe some lines between the fields, or little emojis preceding each field (again I'm totally open to ripping off github)

For now, I'd be OK with sticking to these obscure IDs, we can tackle migrating to user names in a separate PR. 

We should soon also add references to people in the public analysis listing (and review tab somewhere), which links to people's profiles.

adelavega:
I'm also thinking it looks weird to have a profile image place and no way to add one... maybe gravatar is good enough,

adelavega:
Ah, and maybe we can automate linking to ORCID although that's lower priority.

adelavega:
Ah, and I think this is a general note, but let's make it so when the analyis list is empty, it says something more useful than "No Data".

Maybe "No analyses to show" (if this is generic wording), or if its only shown here ("This user has not created any analyses")

adelavega:
Oh, and this is pretty big and probably should be something on the neuroscout side of things, but I think all drafts should be considered "unlisted" (not quite private, you should still be able to access read-only if you have the URL, but not be able to browse them).

adelavega:
I'm actually starting to think that if your profile lists your analyses, maybe this should sort of replace "My analyses", and sort of like in github when I was to browse what I've done, I go to my public profile and do that. But when I'm there I can see everything, and when someone else is there, they can only see my public stuff.

adelavega:
Can you also change the text of "Profile" in the dropdown to "My Profile" and make it the first item in the list? Thanks

adelavega:
@jsmentch if you have issues with the frontend, log in and log out. this is a consequence of loading this new feature.

adelavega:
Hey @tyarkoni this is a PR that adds profile functionality for users. This is currently live on the server if you want to take a look and give some feedback.

Basically, if you go to the top right menu -> My Profile, you can see your public profile, which links you to your public profile.
The profile's address is `https://neuroscout.org/profile/{user_id}`. You can then click on `Edit Profile` to change these fields.

On the right, it shows all of your public analyses. This is the same another user would see. This does not include any DRAFTS or compiled private analyses.

On the edit page, you can choose if you want your email to be publicly viewable or not, as well as edit the value of your profile fields.

We also changed the registration to ask new users for some of these fields (some are new values so old users may not have a default value).

In terms of how to integrate these profiles to the website, a few options/questions:

1) One option is that when browsing "Public Analyses", each analysis lists under neath the user's name, and you can click on it
    Relatedly, we also need to add the author of an analysis when you click on it and are viewing the Summary.
2) Another option is that there could be a public listing of all users. I'm not a huge fan of this, as I don't see the point and it exposes all users, but it's an option.
3) Currently users have an arbitrary sequential ID, which is how we come up with the address for their profile. We could also ask users to create a "username", and/or automatically create one based on their email for existing users. We could then show this instead of their full name, and use it for the profile's URI.

Let me know if you have any other thoughts.

tyarkoni:
Quick thoughts:

* The interface itself looks good and seems intuitive; nice job!
* That said, editing currently seems to be broken, at least for me. When I click update profile, nothing happens (there isn't an error message or anything, it just doesn't do anything).
* Re: integration, I like your option (1). I agree that there isn't a need for a listing of all users.
* Re: user IDs, yeah, it would be good to not sequentially number user IDs... that would very easily allow someone to view/scrape all profiles in order. I realize users still have control over privacy, but mistakes can be made, and it would be better to at least require the viewer to know the username of the person they're looking for. I don't know that we want to mandate creation of a username, but maybe we could just generate a short hash for users by default, much like for analyses. Then that could be an editable field users can change at any time, with the only constraint being that usernames have to be unique.

adelavega:
Okay, that non-editable bug seems like a minor one, already spotted it.

Otherwise, the rest makes sense. We could auto-generate user IDs that are some what human readiable based on their email address, plus a number of that user_id is already taken. Then allow them to edit it to a unique value.

adelavega:
@tyarkoni that bug is fixed. if you want to edit your profile!
Oh, and one more minor thing. Currently we have a place for an avatar but it's only loaded for google users, since we can pull it from a link rather than dealing with uploading avatars.

Another option would be to use gravatar to get profile pics based on email. Any preferences here?

tyarkoni:
No preferences re: avatars.

BTW, the editing interface works now, but when I'm logged out, it doesn't display any of my analyses, even though my account is set to public and some of my analyses are verifiably public.

adelavega:
@rwblair to summarize what else we should do:

- Fix logged out profile viewing bug
- Add author to analysis list (if it fits) but at minimum to analysis itself, and link author name to profile
- Add auto-generated user names (from email), and ability to edit these user names for each user
- Fix tests

adelavega:
I can confirm that when I view my profile publicly, it looks different than when I view it logged in. I still see analyses but not all of them (109 pages vs 4).

I'm confused because I though that the analysis listing for publiic profiles was now pulled from a different route that would only show public ones...

PR #796, opened 2020-07-13
Remove defunct report args

adelavega:
- Remove defunct arguments to generate_report

### DIFF
--- a/celery_worker/tasks.py
+++ b/celery_worker/tasks.py
@@ -10,9 +10,9 @@ def compile(hash_id, run_ids, build):
 
 
 @celery_app.task(name='workflow.generate_report')
-def generate_report(hash_id, report_id, run_ids, sampling_rate, scale):
+def generate_report(hash_id, report_id):
     return report.generate_report(
-        flask_app, hash_id, report_id, run_ids, sampling_rate, scale)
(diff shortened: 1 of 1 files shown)


adelavega:
@jsmentch this should fix your pending report issue. this is up live, can you confirm it fixes your issue?

adelavega:
okay i'm pretty sure this is it so i'm going to go ahead and merge but let me know if this crops up again.

jsmentch:
I'm currently having some issues accessing the site where it will go to a white screen shortly after going to https://neuroscout.org/builder or upon pressing any keys

If I try to go to that previous analysis I receive this error : TypeError: this.props.checkJWT is not a function

adelavega:
sorry that should be fixed. i accidentally loaded a feature that is being worked on

adelavega:
make sure to do a hard refresh if you get another error

jsmentch:
The web interface still is hanging for me when trying to get a new report from that analysis ArqX8. Even in a new incognito window. But, if I make a new analysis wQDJ8 using some of those same uploaded predictors, there is no problem! So it may be okay

adelavega:
What's going on here most likely is that the hung result is 'cached' for that analysis, which explains why it clears up on a new analysis.

PR #799, opened 2020-07-14
If user is loading new empty analysis lets assume they own it.

### LINKED ISSUE #798: read-only warning always shown
@rwblair for some reason the read-only warning is being shown even when i'm the owner of the analysis.

neuroscout.org is currently at master and im seeing this issue.

rwblair:
fixes #798.


adelavega:
LGTM

PR #811, opened 2020-08-15
Speed up and fix My Analyses

adelavega:
/myanalyses is quite slow right now.

One improvement is to only return the fields that matter for the listing, for both private and public analyses.

This seems to reduce the data transfered by 5x and speed up the query by up to 10x. 

However, I'm still getting a situation where the query hangs on the frontend, and I only see a spinning icon. 
I tested this with an account with only a few analyses, so I don't think having too many analyses is the issue.

Instead, I think there's scenarios where the logic goes awry and it just hangs.

### DIFF
--- a/neuroscout/frontend/src/App.tsx
+++ b/neuroscout/frontend/src/App.tsx
@@ -42,7 +42,7 @@ let checkCount = 0;
 
 class JWTChange extends React.Component<JWTChangeProps, {}> {
   jwtChanged = memoize((jwt, user_name) => {
-    if (!!jwt && user_name > 0) {
+    if (!!jwt && !!user_name) {
       this.props.loadAnalyses();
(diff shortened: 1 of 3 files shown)


adelavega:
I'm also seeing scenarios where it says "No Analyses" even when I know that is not the case.

adelavega:
@rwblair let's take a look on monday. I think there's actually something funky with this, and I'm seeing it across accounts, so I don't think it's just my huge analysis list, like I thought earlier.

adelavega:
Unfortunately looks like this is behavior from our recent PR... so you were right @rwblair just had to merge to find the bugs.

I reverted the server to a previous state for now as the "My Analysis" listing is pretty important.

adelavega:
Good to go, app feels snappy.

PR #816, opened 2020-09-11
Update to Marshamallow 3, also upgrade webargs flask_apispec

adelavega:
We're using a fairly old version of Marshmallow, so I'm upgrading to v3, and upgrading `webargs` `flask_apispec` as well.

In the process, I also updated the Dockerfile to buster from stretch.

### DIFF
--- a/neuroscout/Dockerfile
+++ b/neuroscout/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6-stretch
+FROM python:3.6-buster
 ARG DEBIAN_FRONTEND=noninteractive
 
 RUN mkdir -p /usr/src/app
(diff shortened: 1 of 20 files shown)


adelavega:
@rwblair I think I was able to upgrade to marshmallow 3! It wasn't *as* bad as I thought fortunately, just a couple of tricky changes.

Do you mind testing this against the frontend locally on your dev machine? If it looks good I'll test on live server.

adelavega:
Oh, right. The main thing that might still not work is the multi-part file upload. Need to test that out still...

adelavega:
Fixed the file upload by doing a double Marshmallow Schema that ignores extra arguments. This accept all arguments, but still appropriately documents them.

Issue #826, opened 2020-10-08
template latest version on frontend


adelavega:
instead of `latest` the specific version tag.

maybe pull directly from github

rwblair:
This was specifically for version of the docker image used in the command shown in the review tab.

adelavega:
@rwblair looks like we can use the docker hub api to pull tags directly:

https://hub.docker.com/v2/repositories/neuroscout/neuroscout-cli/tags

if we took those results and sorted by date, and excluced "latest" we could get the most recent version of the neuroscout-cli docker image available. 
 
is this something that would be allright to handle directly in the frontend?

rwblair:
Just ran curl on that endpoint, looks good. Can work with it, looks to me like its already coming back sorted by date so hopefully wont have to deal with pagination.

adelavega:
closed by #837

Issue #821, opened 2020-10-07
reports fail if subjects' runs mismatch (pybids limitation) 


jdkent:
Review section (Design Reports)

for the Life dataset, when I added subject rid0000005 run - 2, I got an error on the page:

Report failed to generate
[Error validating analysis, `rms`](https://youtu.be/VWUGOEgAsSQ?t=1326)

don't know if this is just a small idiosyncratic error for this participant or something larger.

adelavega:
The problem is that pybids entities used to require a design matrix are: subject 1 and 5, run 1 and 2, but report is for subject 1 run 1, subject 5, run 2. So only two runs event's are generated, and pybids is expecting 4 runs of events. 

We need a way to specify to pybids to only generate models for those two specific runs, not all combinations of runs and subjects. 

One option would be to loop over runs outside of pybids when generating models / plots.

adelavega:
@rwblair there is also a bug here in which the frontend shows a previous report even though the current report has failed. 

We requested 3 runs, but it failed (due to pybids), however the frontend still shows a previous report (which is for subject 1 run 1 and 2). However, the runs are titled wrong (subject 1 run 1, subject 5 run 2). Some logic which matches the results to the run entities must be failing in this edge case. 


![Screenshot from 2020-10-08 16-03-09](https://user-images.githubusercontent.com/2774448/95513466-050af480-0980-11eb-894e-aeb56d8135c8.png)

adelavega:
I also think we should move the errors to be above the DM (so they dont go unnoticed)

Issue #824, opened 2020-10-07
retype password on sign up


adelavega:
- it's fairly customary and we don't do it
- this can be purely a frontend check

tyarkoni:
+1

Issue #822, opened 2020-10-07
unable to deselect "scale design matrix"


jdkent:
under the review tab, it appears [I was unable to deselect "scale design matrix"](https://youtu.be/VWUGOEgAsSQ?t=1197)

adelavega:
relatedly, correlation matrix seemed to be missing

PR #837, opened 2020-10-14
Various UI improvements + Pybids 0.12.3

### LINKED ISSUE #770: Reports: Throw warning when n/a present
Currently, for BERT surprisal if I don't replace the n/as, the  variable turns to all n/as

### LINKED ISSUE #795: clean up API
Some possible minor improvements to API usability:

- Remove `run_id`s return from `/api/datasets`. It seems unnecessary as usually you want run_ids for a specific task, not the whole dataset.
- Allow `/api/predictor` to be queried without `run_ids` and instead `dataset_id` and/or `task_id`. 
Alternatively, add the following: `/api/dataset/<id>/task/<id>/predictors`

### LINKED ISSUE #823: allow unititled analyses?
should we allow analyses to not have a name? 
names can always be given later, and i often find myself struggling with name when im just trying things out.

maybe we can allow users to create without a name, but a name must be given prior to compliation?

### LINKED ISSUE #833: make hash_id 8 chars and case insensitive
to prevent case related clashes.

adelavega:
- Closes #823 
    - Allow united analyses until compilation (then backend rejects compilation until you provide a name).
-  Closes #833 and #838 
   - Switch to 5 char UUID, checking for clashes upon generation
- Closes #770 by throwing warning if any n/as are found in variables. It suggesting using the Scale transformation to remove n/as.
 - Default option for scale is to replace n/a with 0 after scaling.  
- Closes #795. Minor API improvements.

### DIFF
--- a/celery_worker/requirements.txt
+++ b/celery_worker/requirements.txt
@@ -24,5 +24,5 @@ Flask-Caching==1.3.3
 flask_jwt
 shortuuid==1.0.1
 sentry-sdk==0.13.0
--e git+https://github.com/bids-standard/pybids#egg=pybids
+pybids==0.12.3
 Werkzeug==0.15.3
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,6 +33,7 @@ services:
   postgres:
     restart: always
     build: ./postgres
+    shm_size: 1g
     volumes:
       - /media/neuroscout-data/neuroscout/db_data:/var/lib/postgresql/data
     expose:
--- a/neuroscout/Dockerfile
+++ b/neuroscout/Dockerfile
@@ -22,6 +22,7 @@ RUN npm install -g yarn
 COPY requirements.txt /usr/src/app/
 COPY optional_requirements.txt /usr/src/app
 RUN pip install setuptools==45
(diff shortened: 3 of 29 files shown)


adelavega:
@rwblair can you also tackle #826 on this one?

adelavega:
For the `n/a` problem:
Scale - make replace after scale default value
Add docs in warning about doing this.

adelavega:
Fixed #821 by using new version of pybids which has a fix for this. Also bumped to latest pybids version. Need to update the PIN once the official release comes through.

We should also address the fact that correlation matrix plots are not showing up in this PR.

rwblair:
f02e32e fixes #826

adelavega:
@rwbair just a reminder to tackle #824 tomorrow if you get a chance (in this PR if it's still open, thanks).

adelavega:
Just to recap things left to do:
- Waiting to @tyarkoni to review my pybids PR, so we can cut a new release and PIN it here
- @rwblair we might want to look at this issue as well: #822
- @rwblair working on #824
- I'm also looking into why the corr matrix plot is not showing up.

PR #844, opened 2020-11-10
Graph extraction, across all active datasets, plus downsampling of features

### LINKED ISSUE #768: Add Graphs to basic extractors
Currently only supported for tokenized extractors.

This could allow adding a downsampling filter post-extraction in #596 to prevent too dense predictors (too much disk space is used)

adelavega:
In preparation for extracting audio regressors, this PR adds a few features:

- Graph extraction for basic extraction. Closes #768 
   - Assume that the first node contains a transformer that is the extractor
- Modified the `manage.extract_features` function to default to extracting in all active Datasets/Tasks if no specific Dataset is specified. 
- Resampling can be specified via `extract_features` command or feature_schema

In DB:
- Extracted several auditory features with downsampling (see new feature schema entities)


adelavega:
Let's use resample (as a pliers util) and add logic in Neuroscout for when to add it. We could add a parameter that applies to every Graph `extractor_args` which includes `resample=` and a frequency value (or None, does no resampling). We could set the default value to 3hz so we don't forget or just be strict about adding that. 

Should probably also find a way to log that things were resampled.

adelavega:
Maybe we could append to the Extractor history

adelavega:
A problem is that since we pre-convert Videos to Images, the resample function does not see the whole video at once, only the individual frame, so resampling doesn't work for Images.

One solution would be to use Graphs for the conversion, which would be really slow. 

Or to recreate the VideosFrameStims when extracting these stimuli. This seems like a better option but may also be a bit of a pain. We could also simply extract at 3hz, to at least match the frequency of audio features (even if it was not resampled).

adelavega:
Need to also add the resampling to the Extractor history for provenance sake.

adelavega:
Getting weird results for RMS in Budapest. Looks like rms get extracted at 0.75hz, which is odd. 

Either way, the logic for upsampling may be flawed in pliers, so I also need to take a look at that again. Could potentially only make it "downsample"

PR #849, opened 2020-12-11
Ingest Naturalistic Neuroimaging Database

### LINKED ISSUE #848: ingest naturalistic neuroimaging database
This is a set of issues to take care of to ingest the Naturalistic Neuroimaging Database

cc: @sarahucl94

1) Preprocessed data
- Preprocessed data for this dataset already exists and is available in the OpenNeuro repo, but is not fully BIDS Derivative / fitlins compliant. Adapting this data is the best way forward (vs running in fmriprep), and @sarahucl94 is willing to update the OpenNeuro repo to make it fully compliant. 
- The preprocessed data was concatenated across runs, thus I will need to make a "quasi RAW" dataset that has this same structure to ingest the dataset into Neuroscout
- The filenames of the preproc data need to be renamed to add `space-` and `desc-preproc`. We also need subject specific `-mask` files. It's possible that the MNI group level masks could be copied over for every subject (and/or be available higher up in the hierarchy), since the images are aligned in MNI space already.
- The data are already motion and ICA detrended, so there are no explicit confounds to ingest
- We may have to host the data once the changes are made in `neuroscout-datasets` to make it easier to fetch on run time (or perhaps we can fetch only that part of the raw dataset directly from openneuro... that might be ideal).
- I need to also double check fitlins' requirements for preproc data (e.g. which specific files are needed).

2) Annotations
- @sarahucl94 already has some annotations we can ingest directly into neuroscout, rather than re-extract.
Most importantly, speech, faces, and perhaps later on character encoding. 

3) One or many tasks?
- The dataset has 10 movies, and each subject only saw one movie. Therefore, treating all of them as one "task" (i.e. movie watching) might be ideal, to allow easy analysis of the entire dataset.
- However, we still want to allow ppl to only analyze one task, and the raw dataset has task encoded as different movies. 
- I think ideally we want to change Neuroscout to allow more than 1 task per analysis, and ingest as is (i.e. different tasks).

### LINKED ISSUE #857: record estimator in upload meta-data
AFNI / nilearn in `NeuroVaultUploadCollection`

adelavega:
Closes #848 (ingest NND)
Closes #857 (record estimator used in fitlins in db)
Also addresses #852 on the server side. Now the API should be fully oriented around selection of runs, which can be one task, a combination, or all. 

NND models ran successfully in AFNI, but ran out of memory w/ nilearn due to concatenation of runs into one big image. 

Remaining issue: NND is already smoothed. How would this work for Neuroscout?


adelavega:
Status: 
- Check on AFNI fitlins
- Check that fitlins is not adding confounds
- Check where the memory runs out in nistats process
- Run job in high memory node on ls5

adelavega:
Error in reports:
```
worker_1      |   File "/celery_worker/tasks.py", line 15, in generate_report
worker_1      |     flask_app, hash_id, report_id)
worker_1      |   File "/usr/local/lib/python3.6/site-packages/neuroscout/tasks/report.py", line 122, in generate_report
worker_1      |     analysis, pes, bids_dir, report_obj.runs)
worker_1      |   File "/usr/local/lib/python3.6/site-packages/neuroscout/tasks/utils/build.py", line 92, in build_analysis
worker_1      |     analysis, predictor_events, tmp_dir, run_ids)
worker_1      |   File "/usr/local/lib/python3.6/site-packages/neuroscout/tasks/utils/build.py", line 65, in writeout_events
worker_1      |     entities, path_patterns=PATHS)
worker_1      |   File "/usr/local/lib/python3.6/pathlib.py", line 899, in __truediv__
worker_1      |     return self._make_child((key,))
worker_1      |   File "/usr/local/lib/python3.6/pathlib.py", line 686, in _make_child
worker_1      |     drv, root, parts = self._parse_args(args)
worker_1      |   File "/usr/local/lib/python3.6/pathlib.py", line 640, in _parse_args
worker_1      |     a = os.fspath(a)
worker_1      | TypeError: expected str, bytes or os.PathLike object, not NoneType
```

adelavega:
I'm going to go ahead and merge this, as the results look good.

The resolution re: smoothing is that Jeremy Skipper will provide me with non smoothed images at some point. In the meantime, we can disable smoothing at run time.

Issue #865, opened 2021-01-20
Make tooltips more obvious


adelavega:
Currently the (i) icons are hard to locate. Instead lets make them a bigger button with HTML tooltips with rich documentation

adelavega:
This could possible use a HTML tooltip that only pops up on hover, or if its too annoying requires a click, but is obvious (i.e. is button)

PR #870, opened 2021-02-02
[WIP] Update predictor tab

### LINKED ISSUE #695: improve predictor selection / search tab
- Filter features by extractors (same filter option used for Runs in dataset selection tab)
- add link to pliers API reference for each Extractor (jump directly at extractor)
- maybe add explicit info on modality
- [ ] Come with descriptions of extractors and add to API

rwblair:
fixes #695. Also doing upgrades of frontend packages, most importantly antd 3 to 4.

- [ ]     Fix class names used in cypress test for finding elements
- [x]     Remove description column
- [x]     Adding "+" to show expandable content, which can include the longer descriptions, link to docs, precomputed stats (min/max, contains n/a) etc...
- [x]     Show extractor descriptions when hovering over extractor.
- [x]     Tweak threshold for putting selected predictors below so that table is wider more often
- [x]     Make checkbox narrower
- [x]     Remove interior scroll for the table, and have it rely on solely the pages scroll bar which is already in use on most screen sizes.

Current plan for filters:
- [x]     Have filters exist outside of table, rows of checkboxes for categories to filter on in the right hand column above where the list of all selected predictors show up as tags.

[improve predictor selection / search tab](https://app.gitkraken.com/glo/card/dba060b4fe4d42f6b0913c02322a31fc)


adelavega (changes requested):
Looking good. Here's some initial thoughts:

- I think all the predictors should still load by default, and only when you select a filter on the right should the filtering begin.
- So it would be no check boxes checked = show all. Once one checkbox is shown, then it becomes a filter. I think this is similar to how ebay operates. 
- This also allows full text search to operate by default. 
- Can we show the number of matching items in parentheses after the filter (e.g. ClarifaiAPI (28)). This is like ebay also.
- Given that we're not showing the Predictor `description` any more, and we're short on width, let's make the table less wide and give more space to the filters on the right. 
- Minor text tweak: Predictor Source Filter-> "Source" in bold.. Same for modality. Just "Modality:" in bold. 
-
I'm less sure on this but its worth experimenting:
- I'm starting to think showing selected predictors below would be better. It is not visually distinct enough right now. 
- It might look cleaner to separate the source filter by type. e.g. "Extractor:" "Collection:" "Confounds:"

Bugs:
- Modality filter doesn't seem to be working
- When I selected "Clarifai" and then did a text search for "car" I got 1 result, but I couldn't select it. Actually I can't select anything now, after selecting a single fmriprep confound. After removing that from the list I was able to select again.

adelavega:
I do think it looks too long / messy when a lot of predictors are selected:

image

adelavega:
"Clear all" button would be nice too.

adelavega:
Notes from lab meeting:

    Add selected predictors tab ("shopping cart")
    When you switch to new tab, all things are unchecked (put numbers in parantheses update)
    Remove tags from bottom right
    Parentheses items matching (updates on search)
    All unchecked = all selected
    Default to all unchecked

adelavega:
@rwblair let's make sure to address #865 in this PR as well. I think we could just change the (i) to a question mark in blue. Perhaps a button with <QuestionOutlined /> icon.

adelavega (changes requested):
Here's some quick first impressions. Overall it looks great and I think the design is very close to what I envisioned. Only minor comments and some bugs. 

- Aesthetics looks nicer for some reason, maybe ant upgrade. Nice!

Bugs I found:
- When I selected both first `text` and then and `LengthExtractor`, filtering was not applied for Source. 
- When I select LengthExtractor, then audio, only modality is applied. Seems related to prior point. Multi filtering logic on modality + source seems funky. 


image

^ this happens when I searched for "Speech" and selected it (after already having added 1 thing to my cart). Actually, this seems to happen anytime you do two searches, and when selecting a second item to add. 

These are more general comments / notes to self:
- Can we think of something other than "Overview" for the first tab name. Maybe "Start" and remove header. 
- Change some of he Source names to be more user friendly (e.g. ComplexTextExtractor) as well as modality names to match intuition.

adelavega:
db changes:
- Round off numbers for summary statistics
- Fix minimum missing numbers

frontend changes:
- Add text for n/a hover over (i'll give copy)
- Remove sample frequency from display (on frontend)
- Predictor description be its own line regardless
- If number of `n/a`s is 0 don't display it. 
- Add Extractor description to (+), and hover over "Source" should be Extractor description

adelavega:
Filter changes:
- OR within Source & modality, and AND between source and modality for filtering logic
- Update counts for what is shown in table (i.e. start w/ full counts and as filters applied the numbers get smaller/updated)
- Break up "Source" into 3 categories: "Extractors", "Confounds", "Other"
- Hover overs for Extractor description in filter list

adelavega:
The requested db changes are in progress right now (executing)

rwblair:
Previous two comments as a single task list so I can keep track of whats left to be done:
- [ ]     Add text for n/a hover over (i'll give copy)
- [ ]     Remove sample frequency from display (on frontend)
- [ ]     Predictor description be its own line regardless
- [ ]     If number of n/as is 0 don't display it.
- [ ]     Add Extractor description to (+), and hover over "Source" should be Extractor description
- [ ]     OR within Source & modality, and AND between source and modality for filtering logic
- [ ]     Update counts for what is shown in table (i.e. start w/ full counts and as filters applied the numbers get smaller/updated)
- [ ]     Break up "Source" into 3 categories: "Extractors", "Confounds", "Other"
- [ ]     Hover overs for Extractor description in filter list

adelavega:
Having this issue building:

```

/neuroscout/neuroscout/frontend/node_modules/vega-lite/build/src/index.d.ts
(1,25): Cannot find module '../package.json'. Consider using '--resolveJsonModule' to import module with '.json' extension

```

Issue #874, opened 2021-02-02
Speed up stimulus loading on ingestion


adelavega:

adelavega:
It currently takes up to 30 mins for NND to load all the stimuli.

Current approach is to load all stimuli into pliers Stimuli classes prior to checking if they are compatible with the extractor. This is because it is much easier to check using pliers' internal l ogic, but it is much slower. It would be better if only the necessary stimuli are loaded (although it may still take quite a bit of time for some such as visual stimuli classes)

adelavega:
Might want to also only load when necessary. As is, it has to load all of them into memory which takes up a lot of RAM.

adelavega:
Closed by #875

PR #875, opened 2021-02-03
Improve extraction speed

adelavega:
- Direct inserts instead of `bulk_insert_model` which is much faster and skips ORM alltogethr
- Check `mimetype` before loading models, prevents unnecessary loading


adelavega:
Closes #874

adelavega:
Merging despite flaky frontend tests

PR #879, opened 2021-02-06
Serialize objects early to save memory

adelavega:
To address memory problems with Clarifai Extractor, serializing objects early should save memory


adelavega:
I was able to extract Clarifai labels for the dataset after these optimizations.

PR #891, opened 2021-03-17
Add stimulus timing route

### LINKED ISSUE #829: make stimulus timing available via API for each run
For the feature explorer, jeff requested that if we make available the offset from the acquisition onset, to the stimulus onset, it would help for the app to know when to display the provided video.

adelavega:
Closes #829


adelavega:
@rwblair i'm going to merge this PR as is but see the failing tests as example of flakey front end tests. It's happening very often so it might be worth prioritizing

Issue #902, opened 2021-04-14
add active flag to runs


adelavega:
and only return those in the API by default

PR #903, opened 2021-04-14
Add active column for runs, and fix minor compilation issue

adelavega:
- Add `active` attribute to  Runs to be able to disable them on the API
- Fix sequential compilation of analysis


adelavega:
Close #902

PR #924, opened 2021-05-04
try lazy loading Builder component

rwblair:


adelavega:
Actually, this caused the router to fail (e.g. neuroscout.org/builder). Reverted the PR merge commit on `master`

PR #935, opened 2021-05-12
Improve cypress tests

rwblair:
Haven't been able to reproduce issues locally, but these settings seem like a starting place.


adelavega:
Awesome thanks!

PR #941, opened 2021-05-20
Use run id as a string for its value in report select dropdown.

rwblair:


adelavega:
This fix works!
Tests need to be updated. 

Also including #942 in this PR.

adelavega:
@rwblair since you  mentioned fixing the scale issue was going to be more in depth, then instead let's merge this for just this fix

however, looks like tests are not passing, any clues?

PR #945, opened 2021-06-03
Update frontend build process.

rwblair:
This PR moves us from abandoned react-scripts-ts to mainline react-scripts. Forced to switch using eslint over tslint. Also tightens up our tsconfig, stated using prettier.


rwblair:
@adelavega any idea why last_activity_ip might be set to None in that failing test?

adelavega:
That's weird.

That line is set here: 
`
        current_identity.last_activity_ip = request.remote_addr
` 

everything a user access any route. Honestly, it's not that important but it a weird regression. 

These tests pass locally so disabling this line for now.

rwblair:
@adelavega I re ran the test workflows again after my last push to make sure they still passed. Not sure how to make the flake buddy go green.

I've pushed latest to deepdream if you want to confirm the scaling in the vega embed works how we want it to.

adelavega:
Okay, one minor thing I noticed is the aesthetics changed with this PR.

Compare the spacing on the navigation buttons on deepdream vs production. Slightly smaller spaces on deepdream/staging between the icon and the text. For example on the "browse button". I think this is the case on tags as well. (e.g. "status: PASSED"). The space between the lock and the text is less.

Probably we were using a way to add a space by force that is now not working.

adelavega:
Other issues I found:

- Spacing issue with Author names in `Review` tab 

Screenshot from 2021-06-25 14-27-42

- Add Contrast interface does not work
- Correlation Matrix shows Design Matrix

adelavega:
- Add space next to "New Analysis" button
- Clear Filter button doesn't take into account "Task" (in multi task situations)
- Switching datasets should Clear Filters

PR #970, opened 2021-07-30
Filters and search for analysis table

rwblair:
Also adds search to datasets table in analysis builder overview.


adelavega:
- [ ] Need lowercase search

PR #971, opened 2021-08-03
Dont add runs to bids model if all runids selected.

### LINKED ISSUE #968: Drop runs from BIDS Stats Model if no runs are excluded
A limitation of the BIDS Stats Model is that when subjects have different run labels, it is impossible to say which runs to include for which subjects.

For example, in the the `narratives` dataset, task: `pieman`, some subjects have two runs (number: 1, 2), whereas some subjects only have one run (number: 1) and other subjects only have a run but with no run number.

Neuroscout sets the `runs` object in the `Inputs` section of the StatsModel to `[1, 2]` because it takes the set of all run numbers selected.

However, this will cause loading for subjects with no run numbers to fail, because in `analysis.setup()` the default selectors will come from the model, even for subjects with no run numbers.

I suggest as a quick fix that if we detect that all runs are selected, then `runs` is not included in the `Inputs` section.
(Note we need to do this otherwise fitlins will also choke on execution)

An alternative would be th drop `runs` from `Input` all together, as there are other weird edge case scenarios where there could be clashes. For example, if for subjects 1-10 you use run 1, but then 10-20 you use run 2. 

Since in Neuroscout we only ever include events for the specific runs we select, I think it should be OK to do this.

Thoughts? 

cc: @effigies @rwblair

rwblair:
fixes #968

### DIFF
--- a/neuroscout/frontend/src/analysis_builder/Builder.tsx
+++ b/neuroscout/frontend/src/analysis_builder/Builder.tsx
@@ -326,11 +326,14 @@ export default class AnalysisBuilder extends React.Component<
       .filter(x => x.id === this.state.selectedTaskId)
       .map(y => y.name)
 
-    let runs: number[] = this.state.availableRuns
-      .filter(x => this.state.analysis.runIds.find(runId => runId === x.id))
-      .filter(y => y.number !== null)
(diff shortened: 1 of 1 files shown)


adelavega:
LGTM! Thanks!

PR #1002, opened 2021-10-28
set resolver to dockers internal dns for first 30 seconds

rwblair:
If the swagger containers hostname wouldn't resolve nginx would fail to start. I put in the set directive by its proxypass to get around nginx trying to do the hostname lookup on startup. Nginx was complaining about not having a resolver when you visited /api/swagger. Made change according to this:

https://sandro-keil.de/blog/let-nginx-start-if-upstream-host-is-unavailable-or-down/


adelavega:
^~ vs ^ vs none? otherwise let's merge this so it fixes swagger-ui and we can fix the problem of restarting nginx (when swagger fails) in another PR later.

PR #1007, opened 2021-11-10
Add summary tab

rwblair:


adelavega:
Okay I've got one minor request:

image

@rwblair: Could you add text that says "Predictors:" Prior to the tags? Otherwise it lacks context

adelavega:
Otherwise LGTM

PR #1023, opened 2022-02-08
Predictor details / Dataset & Predictor pages

### LINKED ISSUE #974: make it easier to discover features upon landing
feedback from pestilli lab:

it's not clear what features are available. it would be nice if when landing on neuroscout it was easier to explore the datasets + features available on each.

we've gotten feedback like this before so let's make it happen

rwblair:
Add api endpoint to find all analyses and datasets that have a predictor by the same name. Add frontend page to display it all.

- Add pages for Datasets, Predictors, including listing meta-data and which analyses have been used by each predictor
- Closes #974


rwblair:
- Add dataset table
- Show extractor and link to pliers dock
- More documentation for predictor from api/link to docs
- filter by neurovault
- show author of analysis
- skeleton load for full page, show loading for analysis predictors expansion
- create marshmallow schema for api endpoint

rwblair:
@adelavega changes so far are live on deepdream, let me know what you think.

For "More documentation for predictor from api/link to docs" what exactly should we show for this?

I added a dataset detail page, but I'm not sure what all we want to put on it/how we want it organized. Right now has table of all analyses that use the dataset, and then each of its tasks predictors as a big pile of tags.

adelavega:
Maybe "Available in these Datasets" looks better?

adelavega:
Looks pretty good overall though! Mostly aesthetic comments:

One minor aesthetic comment is alignment on the left side of the page. Visually it looks a bit off to me

Also, the table seems to change sizes depending on the contents

adelavega:
As far as the `Dataset` page, a few comments:

- The name of the Dataset should probably be at the top of the page, along with a description. I would push down all the analsyes and add a little section about the Dataset itself. 
- the predictors listing seems too long to me. maybe we can stick each task in accordion?

adelavega:
@rwblair take a look at the frontend tests. i just managed to fix them. it was actually failing in a normal way, not related to the server, but I added extra waiting just in case.

That said, there's a lot to update in tests relative to all the new features you've added lately.

rwblair:
Clairfi Extractor transport predictor shows up with no modality in all predictors.

adelavega:
- Dataset list on Predictors page
- No analysis list for Datasets
- Minor aesthetic tweaks


In the future we can add a link from Dataset page to the analysis list viewer, pre-populated to filter for that dataset

adelavega:
okay actually need to fix tests then we can merge!

PR #1049, opened 2022-04-25
Fix tour

rwblair:
Not sure why withRouter broke, but I updated Tour to be a proper component wrapped in withRouter and added typing information as necessary.


rwblair:
Because of how steps is conditionally defined by pathname there is a bug if you go through the tour into the builder section, then close the tour,  then restart it via the drop down, current step will be greater than 3 but the steps array is only 4 long when pathname is builder. Need to reset step count on reopen tour, or not redirect to home on starting it.

adelavega:
Awesome, can you deploy to DD?

rwblair:
@adelavega should be up on deepdream

adelavega:
Looks really good.

Two issues:

1) Log in is acting funny (at least on deepdream). It doesn't close the log in window. 
image

2) Minor: If you start an analysis and it telsl you "you sure you want to leave?" and you say "cancel" it'll start the tour right there.

That said, the behavior is sane, because it highlights the top level elements, and then can't start the Builder.
Not horribly bad, but it is odd, so I guess ideally it wouldn't start the tour if you click "Cancel"

adelavega:
Reminder: only changes needed are fixing tests, and changing the color of the flashy tour button.

PR #1050, opened 2022-04-26
Fix crash in review tab for some analyses.

### LINKED ISSUE #1038: Minified error when visiting analysis
It seems to happen only on occasion:

`Error: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7BDescription%7D for the full message or use the non-minified dev environment for full errors and additional helpful warnings.`

rwblair:
Would occur on analyses with descriptions, removed curly brackets wrapping analysis Description in review tab. 

Closes #1038


adelavega:
Great, fixed my issue. Thanks!

PR #1074, opened 2022-08-05
make help in navbar a drop down

rwblair:


adelavega:
Confirmed that cypress tests are passing locally but I'm going to give fixing them here a shot. 

Looks like it's actually failing on the first attempt at something, but the 2nd attempt always fails on a more basic step (our app doesn't like to be re-tested)

adelavega:
I'm going to merge this and continue trying to fix the tests in another branch.

PR #1079, opened 2022-09-01
Fix Cypress & upgrade to v10

adelavega:


rwblair:
@adelavega awesome, thank you for figuring it out.

PR #1082, opened 2022-09-20
WIP: Add admin panel 

adelavega:
Using `Flask-Admin` to implement an administrator panel.

Needs `Flask-Security` in order to be properly to be locked down


adelavega:
Need to modify deployment strategy so that newest image is pulled if available

PR #1083, opened 2022-09-21
Add build and push step to GH Actions

adelavega:
Makes changes to verify we can rebuild the image, and make it easier to deploy pre-built images.

- Adds a new job to GH Actions that build the neuroscout image and pushes to GH Registry.
- Changes the `docker-compose.yml` file such that it pulls from GHCR, and can be specified to be a pr, release version or master.

**To-Do** (not necessarily in this PR)
 - Integrate tests w/ image building. 
I attempted this but it became complicated because of `docker-compose`. I think a good setup would be one job for image build + push (without needing to pass tests for PRs) and in parallel a `docker-compose` build and test, like is done in `neurostore`. 
Otherwise, I think integrating pushing to GHCR and using that image with `docker-compose` is not clear to me. 

(Note: It may be possible if we add a custom GHA only .yml which pulls the pre-loaded image, that would have a different name)

 - Modify deepdream and production deployment to use pre-built images, and pull / delete old images as necessary


adelavega:
I think I'll leave tests as is, as its much faster to run tests without building the entire image

adelavega:
The one issue is these don't take variables. 
Currently I'm thinking we can stick with git hooks for deployment but with the following changes:

- Both STAGING and PRODUCTION default to pull pre-built images
- Environment variable determines which pre-built image is used. "stable" or equivalent tag can be used. In STAGING, one can always change this variable to test a new build. (Need to check if images are auto-pulled if updated)
- STAGING machine defaults to building frontend changes in addition to doing up & down.
- PRODUCTION machine defaults to using frontend changes pre-built in image to minimize downtime

For now we can test this working in STAGING by pushing by the user to the server. If this works, we can then add a way to trigger GH Action to push to staging given some tag. If that all works, we can then set up this working on production.

PR #1085, opened 2022-09-21
Add production mode and trigger deployment to staging

adelavega:
Picks up where #1083 left off:

In all cases, must specify two files, such as :

`docker-compose -f docker-compose.yml -f docker-compose.dev.yml up -d`

In the following, the local directory is mapped to `/neuroscout` where `gunicorn` runs

- `docker-compose.dev.yml`: Builds image locally, and also maps port `4000` to allow `yarn start`
- `docker-compose.build.yml`: Builds image locally
- `docker-compose.image.yml`: Pulls image from GHCR, with environment variable `${IMAGE_TAG}` setting the TAG (i.e. `master`, a version, or a pull request).

In `docker-compose.dev.yml`: we run off `/usr/src/app` which is pre-built including the frontend. The image is also fixed to `master`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,24 @@
+on:
+  workflow_dispatch:
+
+jobs:
+  deploy-staging:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install SSH key
+        uses: shimataro/ssh-key-action@v2
+        with:
+          key: ${{ secrets.DEEPDREAM_PRIVATE_KEY }}
+          known_hosts: 'just-a-placeholder-so-we-dont-get-errors'
+          if_key_exists: fail
+
+      - name: Adding Known Hosts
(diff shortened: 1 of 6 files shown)


adelavega:
TODO: Find a way to pre-build the frontend while respecting the local config files

- One option is to build the image on the server again, but leave the GHA in here to check the image process works
- Another option is to inject the variables on build using GH Secrets (defaulting to some sensible defaults in case they are not provided)

adelavega:
I added a github action that allows a manual push to the staging server.

In the future we can modify this to be production server, and for this action to occur automatically given certain scenarios.
The one remaining question for me is how to ensure this only triggers after the latest image is built, to ensure that is the image that is pushed and loaded.

adelavega:
Looks like its working but remaining questions:

- Can we make the frontend config that is built within the `Dockerfile` configurable?  Current its fixed to the production variables, but in the staging we re-build the frontend to match the current code. Can environment variables be loaded during the image build process?
- If build mode is to build on current code: does it re-build every time or only if it detects changes? Or only on manual rebuild?

PR #1094, opened 2022-11-23
EXT: TFHubAudioExtractor- SPICE pitch detection

adelavega:


adelavega:
Looks like its working, Will fix front end tests in separate PR

PR #1107, opened 2023-06-30
FIX: Datalad installation path

adelavega:
- Also fix frontend test config


adelavega:
Frontend tests are still failing but I'll leave that for another PR

